PR #1, opened 2019-08-27
feat(operator/x509): New service for x509

german-muzquiz:
* Created `ChangeDetector` interface to split logic of detecting changes.
* Created `ChangeDetector` and `Transformer` to add support for exposing gate with x509 support on its own Kubernetes service.


pkg/deployer/changedetector/x509.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (g *x509ChangeDetectorGenerator) NewChangeDetector(client client.Client, log logr.Logger) (ChangeDetector, error) {
+	return &x509ChangeDetector{client: client, log: log}, nil
+}
+
+// IsSpinnakerUpToDate returns true if there is a x509 configuration with a matching service
+func (ch *x509ChangeDetector) IsSpinnakerUpToDate(spinSvc *spinnakerv1alpha1.SpinnakerService, config runtime.Object, hc *halconfig.SpinnakerConfig) (bool, error) {
+	// ignore error as default.apiPort may not exist
+	apiPort, _ := hc.GetServiceConfigPropString("gate", "default.apiPort")
+	if apiPort == "" {
```
ncknt: If there's no apiPort (but still with `expose` option being set), we may want to remove a `Service` previously set for the x509.

german-muzquiz: Maybe better to check if no apiPort, and a service named `spin-gate-x509` exists

ncknt: Sounds good to me. Also, the whole check should only happen if `expose.type == LoadBalancer`.

PR #4, opened 2019-09-03
chore(config): Use profiles key for service profiles

ncknt:


pkg/halconfig/populate_test.go
```diff
@@ -26,9 +26,20 @@ func TestParseConfigMap(t *testing.T) {
 name: default
 version: 1.14.2
 `,
-			"profiles__gate-local.yml": "test:\n  deep: abc",
-			"profiles__orca-local.yml": "test.other: def",
-			"files__somefile":          "test3",
+			"profiles":        "gate:\n  test:\n    deep: abc\norca:\n  test.other: def",
```
german-muzquiz: How would it look like a ConfigMap definition with profiles? I think Kubernetes doesn't allow them to look like this:
```
apiVersion: v1
data:
  profiles:
    gate:
      test: 1
```
i.e. everything under `profiles` should be a single string. But then for migrating from halyard to operator, creating a single ConfigMap with all the files could get complicated.

How about having separate configmaps for the main halyard config, the profiles and service-settings? If we have separate configmaps then migration would be easier, because you can create a full ConfigMap with all files in a single directory. Then migration could be as simple as:

```
kubectl create configmap spin-mainconfig --from-file=~/.hal/config
kubectl create configmap spin-profiles --from-file=~/.hal/default/profiles/
kubectl create configmap spin-servicesettings --from-file=~/.hal/default/service-settings/
```

ncknt: In this case profiles is a `string` not a map. The code above matches:
```
apiVersion: v1
data:
  profiles: |
    gate:
      test:...
```

german-muzquiz: After talking with Nico, the new ConfigMap would be generated by a new endpoint in Halyard instead of manually with `kubectl create configmap` to account for other differences between traditional Hal configs and the new confgimap contents.

PR #8, opened 2019-09-09
feat(interface): Use interfaces for SpinnakerService

ncknt:
- Transformers get a name
- SpinnakerConfig is added to the transformer instead of as param
- Validators, Transformers, and ChangeDetectors use interface

### DIFF
--- a/pkg/deployer/deploy.go
+++ b/pkg/deployer/deploy.go
@@ -46,7 +46,7 @@ func NewDeployer(m manifestGenerator, c client.Client, r *kubernetes.Clientset, 
 	}
 }
 
-func (d *Deployer) IsSpinnakerUpToDate(svc *spinnakerv1alpha1.SpinnakerService, config runtime.Object, hc *halconfig.SpinnakerConfig) (bool, error) {
+func (d *Deployer) IsSpinnakerUpToDate(svc spinnakerv1alpha1.SpinnakerServiceInterface, config runtime.Object, hc *halconfig.SpinnakerConfig) (bool, error) {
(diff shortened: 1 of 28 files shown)


pkg/deployer/deploy.go
```diff
@@ -135,9 +137,9 @@ func (d *Deployer) commitConfigToStatus(ctx context.Context, svc *spinnakerv1alp
 	status.LastConfigurationTime = metav1.NewTime(time.Now())
 	// gate and deck status url's are populated in transformers
 
-	s := svc.DeepCopy()
-	s.Status = *status
+	//s := svc.DeepCopy()
+	//s.Status = *status
```
german-muzquiz: Can these couple lines be deleted now that they are commented?

pkg/deployer/transformer/transformer.go
```diff
@@ -23,52 +23,52 @@ func init() {
...
+	TransformManifests(scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig) error
 }
 
-// baseTransformer extends Transformer adding convenience methods.
-type baseTransformer interface {
-	transformServiceManifest(svcName string, svc *corev1.Service, hc *halconfig.SpinnakerConfig) error
-	transformDeploymentManifest(deploymentName string, deployment *v1beta1.Deployment, hc *halconfig.SpinnakerConfig) error
+// BaseTransformer extends Transformer adding convenience methods.
+type BaseTransformer interface {
+	transformServiceManifest(svcName string, svc *corev1.Service) error
```
german-muzquiz: If `BaseTransformer` is now public, shouldn't its methods be also public so that code outside this package can implement them?

PR #11, opened 2019-09-10
imp(context): Use proper context for all deployment activities

ncknt:
- Scaffolding for secrets decryption

### DIFF
--- a/pkg/controller/spinnakerservice/spinnakerservice_controller.go
+++ b/pkg/controller/spinnakerservice/spinnakerservice_controller.go
@@ -3,6 +3,7 @@ package spinnakerservice
 import (
 	"context"
 	"github.com/armory/spinnaker-operator/pkg/halconfig"
+	"github.com/armory/spinnaker-operator/pkg/secrets"
 
 	spinnakerv1alpha1 "github.com/armory/spinnaker-operator/pkg/apis/spinnaker/v1alpha1"
 	deploy "github.com/armory/spinnaker-operator/pkg/deployer"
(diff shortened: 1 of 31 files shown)


pkg/controller/spinnakerservice/spinnakerservice_controller.go
```diff
@@ -34,8 +35,8 @@ func Add(mgr manager.Manager) error {
 }
 
 type deployer interface {
-	IsSpinnakerUpToDate(svc spinnakerv1alpha1.SpinnakerServiceInterface, config runtime.Object, hc *halconfig.SpinnakerConfig) (bool, error)
-	Deploy(svc spinnakerv1alpha1.SpinnakerServiceInterface, scheme *runtime.Scheme, config runtime.Object, hc *halconfig.SpinnakerConfig) error
+	IsSpinnakerUpToDate(ctx context.Context, svc spinnakerv1alpha1.SpinnakerServiceInterface, config runtime.Object, hc *halconfig.SpinnakerConfig) (bool, error)
+	Deploy(ctx context.Context, svc spinnakerv1alpha1.SpinnakerServiceInterface, scheme *runtime.Scheme, config runtime.Object, hc *halconfig.SpinnakerConfig) error
```
german-muzquiz: Maybe we can have another struct act as a placeholder for common objects that need to be passed everywhere, like `context` and `SpinnakerServiceInterface`. In this way we could reduce the number of parameters on the functions.
Also `SpinnakerConfig` is created from `config runtime.Object`, maybe we can include the runtime object as part of the `SpinnakerConfig` struct?

ncknt: All parameters have a good reason to be here.  `context` is quite standard and expected. I actually like `SpinnakerConfig` and `config` separated as you can use the former without the latter. My comment on the signature was more for the transformers, change detectors and validators to make it easier for people to contribute. We may address that via good examples and documentation. I'll keep track of that in a separate issue.

PR #13, opened 2019-09-11
feat(operator/doc): Pointers to structs

german-muzquiz:


pkg/apis/spinnaker/v1alpha1/interface.go
```diff
@@ -8,9 +8,9 @@ import (
 type SpinnakerServiceInterface interface {
 	v1.Object
 	runtime.Object
-	GetStatus() SpinnakerServiceStatus
-	GetExpose() ExposeConfig
-	GetSpinnakerConfig() SpinnakerFileSource
+	GetStatus() *SpinnakerServiceStatus
+	GetExpose() *ExposeConfig
```
ncknt: I don't think we want to allow modification of the config. Status yes but not config or expose.

german-muzquiz: Right, config is coming from manifest files.

PR #20, opened 2019-09-19
feat(expose): Specify port to use for exposing in CRD

german-muzquiz:
Change to add a `Port` override to spinnaker CRD expose configuration.
Example:

```
apiVersion: spinnaker.io/v1alpha1
kind: SpinnakerService
metadata:
  name: spinnaker
spec:
  spinnakerConfig:
    configMap:
      name: spinconfig-v001
  expose:
    type: service
    service:
      type: LoadBalancer
      publicPort: 1111
      annotations:
        "service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "http"
      overrides:
        gate:
          publicPort: 7777
```

Also default expose ports are set to `80` for all services (deck, gate and gate-x509).


pkg/util/runtime.go
```diff
@@ -110,3 +112,31 @@ func GetPort(aUrl string, defaultPort int32) int32 {
 	}
 	return defaultPort
 }
+
+func GetDesiredExposePort(ctx context.Context, svcName string, hc *halconfig.SpinnakerConfig, spinSvc spinnakerv1alpha1.SpinnakerServiceInterface) int32 {
```
ncknt: It's exported, can you add a comment?

deploy/crds/spinnaker_v1alpha1_spinnakerservice_crd.yaml
```diff
@@ -59,10 +59,16 @@ spec:
...
                               type: string
                             type: object
+                          port:
+                            format: int32
+                            type: integer
                           type:
                             type: string
                         type: object
                       type: object
+                    port:
```
kevinawoo: i kind of feel like having the global port isn't really worthwhile, and might add confusion.

What do you think about not including it?

deploy/crds/spinnaker_v1alpha1_spinnakerservice_crd.yaml
```diff
@@ -59,10 +59,16 @@ spec:
                             additionalProperties:
                               type: string
                             type: object
+                          port:
```
kevinawoo: what do you think about adding clarity on this by changing `port` to `exposedPort` or maybe `publicPort` or.... hmmmmmmmmm....... I guess `port` is ok?

kevinawoo:
OH MAN, THAT'S A LOT OF CODE.

THANKS! 🙇

pkg/deployer/transformer/x509.go
```diff
@@ -77,9 +77,10 @@ func (t *x509Transformer) TransformManifests(ctx context.Context, scheme *runtim
 func (t *x509Transformer) createX509Service(apiPort int32, gateSvc *corev1.Service) (*corev1.Service, error) {
 	x509Svc := gateSvc.DeepCopy()
 	x509Svc.Name = util.GateX509ServiceName
+	publicPort := t.getPublicPort(int32(80))
```
ncknt: should default to https/443. This is X509 so TLS.

deploy/crds/spinnaker_v1alpha1_spinnakerservice_crd.yaml
german-muzquiz: Having a global port can help to have all exposed services running on the same public port (if there's no `overrideBaseUrl` set), also I think it makes sense to have `overrides` section replace values in the global section.

deploy/crds/spinnaker_v1alpha1_spinnakerservice_crd.yaml
german-muzquiz: `publicPort` sounds good

pkg/deployer/changedetector/expose_lb.go
```diff
@@ -121,3 +126,19 @@ func (ch *exposeLbChangeDetector) exposeServiceTypeUpToDate(serviceName string,
...
+
+func (ch *exposeLbChangeDetector) exposePortUpToDate(ctx context.Context, serviceName string, spinSvc spinnakerv1alpha1.SpinnakerServiceInterface, svc *corev1.Service, hc *halconfig.SpinnakerConfig) (bool, error) {
+	rLogger := ch.log.WithValues("Service", spinSvc.GetName())
+	if len(svc.Spec.Ports) < 1 {
+		rLogger.Info(fmt.Sprintf("No exposed port for %s found", serviceName))
+		return false, nil
+	}
+	formattedServiceName := serviceName[len("spin-"):]
+	desiredPort := util.GetDesiredExposePort(ctx, formattedServiceName, hc, spinSvc)
+	if desiredPort != svc.Spec.Ports[0].Port {
```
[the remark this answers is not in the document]
german-muzquiz: I think it's better to look for the port name that the operator creates than fail if len(Ports) != 1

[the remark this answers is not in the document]
german-muzquiz: Also port len check is done a few lines above

PR #25, opened 2019-10-05
feat(validations): Validation framework

german-muzquiz:
New:
1. Reverse tunnel setup for being able to debug admission webhook locally (run `make reverse-proxy`).
2. Package `validate.configfinder`: Used to load accounts from various sources, goal is to have one go module for each possible config source (main hal config file, profile files, etc.)
3. Use generators to build validators.
4. Run validators in parallel.
5. Register admission webhook to also trigger on `create`, `update` of ConfigMaps, not just SpinnakerService instances.
6. Moved generic test helper functions to new `test` module.

Open issues:
1. How to run validations only once if webhook is triggered separately by each file being applied (SpinnakerService and ConfigMap)?
2. There is a double trigger when validations fail inside a goroutine, needs investigation.


pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -77,6 +83,36 @@ func Add(m manager.Manager) error {
 	return as.Register(validatingWebhook)
 }
 
+func getSpinnakerServiceRule() admissionregistrationv1beta1.RuleWithOperations {
+	gv := v1alpha1.SchemeGroupVersion
```
ncknt: This needs to be pluggable the same way that we use `SpinnakerServiceInterface` on the service controller.

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -35,17 +38,20 @@ var log = logf.Log.WithName("spinvalidate")
 
 // Add adds the validating admission controller
 func Add(m manager.Manager) error {
-	ns, err := k8sutil.GetOperatorNamespace()
-	if err != nil {
-		return err
+	ns := os.Getenv("WATCH_NAMESPACE")
```
ncknt: Why `WATCH_NAMESPACE` here? When would the validation service be in a different namespace than the operator's?

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -92,10 +128,15 @@ func (v *spinnakerValidatingController) Handle(ctx context.Context, req types.Re
...
-	if err := validate.Validate(svc, opts); err != nil {
-		log.Error(err, "SpinnakerService validation failed", "metadata.name", svc)
-		return admission.ErrorResponse(http.StatusBadRequest, err)
+	log.Info("Starting validation")
+	errors := validate.Validate(svc, opts)
+	if len(errors) > 0 {
+		for _, e := range errors {
+			log.Error(e, fmt.Sprintf("SpinnakerService validation failed: %s", e.Error()), "metadata.name", svc)
+		}
+		return admission.ErrorResponse(http.StatusBadRequest, errors[0])
```
ncknt: I'd rather have a new `error` with the messages aggregated. This is what the user will see as a result of `kubectl`, they shouldn't have to dig through the logs.

pkg/validate/account_kubernetes.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (a *kubernetesAccount) GetName() string {
+	p := a.Config.(map[interface{}]interface{})
+	return p["name"].(string)
+}
+
+func (a *kubernetesAccount) GetChecksum() string {
+	p := a.Config.(map[interface{}]interface{})
+	return p["name"].(string)
```
ncknt: Checksum is supposed to be the checksum of all fields that make the config not just `name`. e.g. when you change `namespaces` etc.

pkg/validate/account_kubernetes.go
```diff
@@ -0,0 +1,63 @@
+package validate
+
+import (
+	"fmt"
+	"github.com/armory/spinnaker-operator/pkg/apis/spinnaker/v1alpha1"
+	"github.com/armory/spinnaker-operator/pkg/halconfig"
+	"github.com/armory/spinnaker-operator/pkg/validate/configfinder"
+)
+
+type kubernetesAccount struct {
```
ncknt: I'd expect `kubernetesAccount` to be more typed than `Config: interface{}`. Also can you export so it can be reused in account CRDs?

pkg/validate/configfinder/configfinder.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (f *composedConfigFinder) GetAccounts(provider string) (map[string]interface{}, error) {
+	accounts := map[string]interface{}{}
+	for _, f := range f.Finders {
+		as, err := f.GetAccounts(provider)
+		if err != nil {
+			return nil, err
+		}
+		for name, account := range as {
```
ncknt: This is not how accounts work atm. The last config wins all the accounts. However, I'd like CRD to be additive:

(List1 in `config`, List2 in `clouddriver-local`, List3 as CRDs) -> List2 + List3.

pkg/validate/validate.go
```diff
@@ -2,39 +2,164 @@ package validate
...
 }
 
 type Options struct {
 	Ctx    context.Context
 	Client client.Client
 	Req    types.Request
+	Log    logr.Logger
+}
+
+type Account interface {
```
ncknt: This doesn't seem to be used anymore. The intent of that struct was to provide common utils for account based validation. Account A is validated -> store result with checksum in the status to avoid over-validating.

pkg/validate/validate.go
```diff
@@ -2,39 +2,164 @@ package validate
...
 )
 
-// Transformers tracks the list of transformers
-var Validators []SpinnakerValidator
+// generators is used to register all ValidatorGenerator
+var generators []ValidatorGenerator
 
 func init() {
-	Validators = append(Validators, &singleNamespaceValidator{})
+	generators = append(generators, &singleNamespaceValidatorGenerator{}, &kubernetesAccountValidatorGenerator{})
```
ncknt: We need to be able to call validation given a type of provider. So it'd be nice to have a provider Type <-> validator generator somewhere.

pkg/validate/validate.go
```diff
@@ -2,39 +2,164 @@ package validate
...
+		}
+		for _, v := range va {
+			validators = append(validators, v)
+		}
+	}
+	return validators, nil
+}
+
+// splitSequentialAndParallel returns sorted validators to run in sequence, and the ones that can run in parallel
+func splitSequentialAndParallel(validators []SpinnakerValidator) ([]SpinnakerValidator, []SpinnakerValidator) {
```
ncknt: For readability, it would also be nice to compose the validation graph. Something like:

```
validators = []Validator{
   &singleNamespaceValidator{},
   &parallelValidator{
      &kubernetesAccountValidator{},
      ....
  }
}
```

In that case, we don't need the generator but need to pass the options to `SpinnakerValidator.Validate()`. It'd make adding new validators easier IMO.

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
german-muzquiz: This was used to be able to run the admission webhook locally, taking the namespace from that variable instead of loading it from a file that exists in the pod. What could be a better way to get the namespace when running outside the pod?

pkg/validate/account_kubernetes.go
german-muzquiz: Yeah, I was thinking in checksum implementation to be done in the user story that actually implements the validation, rather than in this one that prepares the code to add validations. What do you think?

pkg/validate/account_kubernetes.go
german-muzquiz: I was thinking in growing typing organically when we absolutely need a field for something, for the reason that account config fields are already defined in Halyard codebase, and if they need to be changed (like in adding a new field), then we'll end up with two places to remember to update instead of one.

pkg/validate/configfinder/configfinder.go
german-muzquiz: Last one wins, but only if the same account name is found in multiple sources. Here the finders are defined in order: https://github.com/armory/spinnaker-operator/pull/25/files/6c11ff601b7617711ab38e633e63b40ff0c8642c#diff-5926d012fef0c048be10851eff01f80dR20
For this example:
```
halconfig:
- kube-1
  ....
- kube-2
  ....
cloud driver-local.yml:
- kube-1
CRD:
- kube-3
  ....
```
The resulting accounts would be:
- kube-1 (from clouddriver-local.yml)
- kube-2 (from halconfig)
- kube-3 (from CRD)

pkg/validate/validate.go
german-muzquiz: You don't know how many `kubernetesAccountValidator` instances you'll need until runtime (one validator per account), so I think generators are still needed.
But I see the value in having the above kind of priority definition with generators instead of using the `GetPriority` approach. With `GetPriority` though, you can fine tune a specific account to be validated before the others instead of at the same time, but I don't know if that's a real use case. Do you want me to change it?

Makefile
```diff
@@ -108,10 +109,19 @@ run-dev:
...
 debug:
 	OPERATOR_NAME=local-operator \
     WATCH_NAMESPACE=operator \
 	dlv debug --headless  --listen=:2345 --headless --log --api-version=2 cmd/manager/main.go -- \
-	--kubeconfig ~/.kube/config --disable-admission-controller
-
+	--kubeconfig ${KUBECONFIG} --disable-admission-controller
+
+.PHONY: reverse-proxy
+reverse-proxy:
```
ncknt: This is cool. I'd rather have it run with a `kubectl run` though instead of having a deployment.

ncknt: Also it doesn't work unless you have a spinnaker-operator service already in there. Maybe we can split the logic between `make setup-reverse-proxy` and `make debug`?

german-muzquiz: I tried with a `kubectl run` but for some reason the deployment kept getting killed after sending that command to the background (and we would now end up with 2 forked processes to manage instead of 1). Also since the pod needs to mount your local ssh public key, the way to specify that in the run command line is with a `--overrides` parameter, which is basically the json spec of the deployment specifying the volume configuration, making the command really long.

Regarding `service` part that is automatically created by the webhook controller setup during startup, there's no need to provision it separately.

Makefile
```diff
@@ -53,9 +54,15 @@ build-dirs:
 	@echo "Creating build directories ${BUILD_DIR}"
 	@mkdir -p $(BUILD_DIR)
 
+# Regenerates CRD yamls out of any changes in spinnakerservice_types.go
```
ncknt: I'd remove that. We have `make k8s openapi` now

pkg/controller/spinnakervalidating/gvk.go
```diff
@@ -18,7 +19,7 @@ func isSpinnakerRequest(req admission.Request) bool {
 func (v *spinnakerValidatingController) getSpinnakerService(req admission.Request) (v1alpha2.SpinnakerServiceInterface, error) {
 	if isSpinnakerRequest(req) {
 		svc := SpinnakerServiceBuilder.New()
-		if err := v.decoder.Decode(req, svc); err != nil {
+		if err := json.Unmarshal(req.AdmissionRequest.Object.Raw, svc); err != nil {
```
ncknt: Why is this needed? It assumes the raw is always json. It's mostly true but rules out protobuf.

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -30,19 +37,129 @@ var log = logf.Log.WithName("spinvalidate")
...
+	}
+
+	// Register webhook server
 	hookServer := m.GetWebhookServer()
-	hookServer.Register("/validate-v1alpha1-spinnakerservice", &webhook.Admission{Handler: &spinnakerValidatingController{}})
-	return nil
+	hookServer.CertDir = c.certDir
+	hookServer.Port = port
+	gv := SpinnakerServiceBuilder.GetGroupVersion()
+	path := "/validate-v1alpha2-spinnakerservice"
```
ncknt: Can you use the gvk to form the path?

pkg/controller/spinnakervalidating/gvk.go
german-muzquiz: `Decoder` is not injected anymore in the new version of the SDK. Good point about protobuf, I'll look for a different way to get a decoder.

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -30,19 +38,129 @@ var log = logf.Log.WithName("spinvalidate")
...
+			return "", "", fmt.Errorf("unable to determine operator namespace. Error: %s and ADMISSION_PROXY_NAMESPACE env var not set", err.Error())
+		}
+		ns = envNs
+	}
+	return ns, name, nil
 }
 
-// spinnakerValidatingController adds an annotation to every incoming pods.
+func deployWebhookService(ns string, name string, port int, client *kubernetes.Clientset) error {
+	// Always recreate the service
```
ncknt: We should update instead of recreating. Also using the manager's client would be preferable to the raw client.

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -30,19 +38,129 @@ var log = logf.Log.WithName("spinvalidate")
...
-	_, err := k8sutil.GetOperatorNamespace()
+	// Determine environment
+	ns, name, err := getOperatorNameAndNamespace()
 	if err != nil {
 		return err
 	}
 
+	// Create Kubernetes service for listening to requests from API server
+	rawClient := kubernetes.NewForConfigOrDie(m.GetConfig())
+	port := 9876
```
ncknt: Can you use a constant for that?

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -30,19 +38,129 @@ var log = logf.Log.WithName("spinvalidate")
...
+	}
+
+	// Register webhook server
 	hookServer := m.GetWebhookServer()
-	hookServer.Register("/validate-v1alpha1-spinnakerservice", &webhook.Admission{Handler: &spinnakerValidatingController{}})
-	return nil
+	hookServer.CertDir = c.certDir
+	hookServer.Port = port
+	gv := SpinnakerServiceBuilder.GetGroupVersion()
+	path := fmt.Sprintf("/validate-%s-spinnakerservice", gv.Version)
```
ncknt: Use the `kind` from `SpinnakerServiceBuilder`. I'd just use from the controller-runtime's webhook.go:
```
func generateValidatePath(gvk schema.GroupVersionKind) string {
	return "/validate-" + strings.Replace(gvk.Group, ".", "-", -1) + "-" +
		gvk.Version + "-" + strings.ToLower(gvk.Kind)
}
```

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -30,19 +38,129 @@ var log = logf.Log.WithName("spinvalidate")
...
+			},
+		},
+	}
+	_, err := client.CoreV1().Services(ns).Create(service)
+	return err
+}
+
+func deployValidatingWebhookConfiguration(configName, ns string, webhook v1beta1.Webhook, client *kubernetes.Clientset) error {
+	// Always recreate the configuration
+	_ = client.AdmissionregistrationV1beta1().ValidatingWebhookConfigurations().Delete(configName, &metav1.DeleteOptions{})
```
ncknt: Same comment as above, we should just update the object

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -30,19 +38,129 @@ var log = logf.Log.WithName("spinvalidate")
...
+				Name:      operatorName,
+				Path:      &path,
+			},
+			CABundle: c.signingCert,
+		},
+		Rules: []v1beta1.RuleWithOperations{getSpinnakerServiceRule()},
+	}
+}
+
+func getSpinnakerServiceRule() v1beta1.RuleWithOperations {
```
ncknt: don't really need that function, can be inlined in `getWebhookConfig`

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -72,7 +194,23 @@ func (v *spinnakerValidatingController) InjectClient(c client.Client) error {
 }
 
 // InjectDecoder injects the decoder.
-func (v *spinnakerValidatingController) InjectDecoder(d admission.Decoder) error {
+func (v *spinnakerValidatingController) InjectDecoder(d *admission.Decoder) error {
 	v.decoder = d
 	return nil
 }
+
+func (v *spinnakerValidatingController) collectErrors(results []validate.ValidationResult) error {
```
ncknt: We're losing errors by just collecting the message. I'd think `ValidationResult` can return a `validationError` that can wrap one or more errors. At the end of parallel validations, we can `ValidationResult.merge([]ValidationResult)` and keep the wrapped errors. We have at least 2 parallel validations: all providers/ci/... and for each provider parallel for each account.

pkg/controller/spinnakervalidating/tls.go
```diff
@@ -0,0 +1,195 @@
...
+	signingCert []byte
+	certDir     string
+}
+
+func getCertContext(operatorNamespace string, operatorServiceName string) (*certContext, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return nil, err
+	}
+	certDir := filepath.Join(home, "spinnaker-operator-certs")
```
ncknt: It seems more standard to take `--tls-key` --tls-cert` parameters (that default to `/etc/certs/tls.key|crt`. If the files are not found they are auto-generated, otherwise people can mount their own cert to these locations and potentially override them via the parameters in the deployment.

pkg/inspect/getter.go
```diff
@@ -43,6 +43,19 @@ func GetObjectPropString(ctx context.Context, obj interface{}, prop string) (str
 	return "", fmt.Errorf("%s is not a string, found %s", prop, c.Kind().String())
 }
 
+func GetObjectArray(obj interface{}, prop string) ([]interface{}, error) {
+	v, err := getObjectProp(obj, prop)
+	if err != nil {
+		return nil, err
+	}
+	var result []interface{}
+	for i := 0; i < v.Len(); i++ {
```
ncknt: That will panic with the wrong params. Test the `Kind()` and if slice, array, then proceed.

pkg/validate/configfinder/configfinder.go
ncknt: AFAIK Spring doesn't reconcile arrays. So in your example, you'd only get `kube-1`. For CRD though, I'd like to allow some accounts defined statically in Spring world + additional ones via CRD.

pkg/validate/parallel.go
```diff
@@ -0,0 +1,82 @@
...
+	return p.aggregateResults(results)
+}
+
+func (p *ParallelValidator) validateAccountsInParallel(accounts []Account, options Options, f func(Account, Options) ValidationResult) ValidationResult {
+	options.Log.Info(fmt.Sprintf("Running validation of %d accounts in parallel", len(accounts)))
+	if len(accounts) == 0 {
+		return ValidationResult{}
+	}
+	resultsChannel := make(chan ValidationResult, len(accounts))
+	abortSignal := make(chan bool)
```
ncknt: How can this be called? In our case we want to stop the validation if `failFast` is true.

pkg/deployer/transformer/testdata/input_deployment.yml
```diff
@@ -0,0 +1,92 @@
...
+    moniker.spinnaker.io/application: \spin\
+    moniker.spinnaker.io/cluster: \gate\
+  creationTimestamp: 2019-07-16T14:13:22Z
+  generation: 20
+  labels:
+    app: spin
+    app.kubernetes.io/managed-by: halyard
+    app.kubernetes.io/name: gate
+    app.kubernetes.io/part-of: spinnaker
+    app.kubernetes.io/version: 2.15.1-rc406
```
ncknt: `1.15.1`?

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
german-muzquiz: There's no `kind` in `SpinnakerServiceBuilder`, just a `GroupVersion` object

pkg/controller/spinnakervalidating/tls.go
german-muzquiz: We can make configurable `certDir`, but the sdk has these restrictions:
* Cert and Key files must be on the same dir
* Cert and Key file names must be named `tls.crt` and `tls.key`, these names are hardcoded in the sdk

pkg/validate/parallel.go
german-muzquiz: When there's a fatal error, channels `resultsChannel` and `abortSignal` are closed (https://github.com/armory/spinnaker-operator/pull/25/files/2b551dd53a23a767a72a62cb88b28fb91a25a5c6#diff-ce77105330da5c799c1121e94a8ded20R49), abort channel is used to simply discard the result of any concurrent validator after it finishes. Exposing this channel to stop validations externally can be implemented in a separate user story

pkg/validate/account_kubernetes.go
german-muzquiz: This will change with the merge with account crd work

PR #29, opened 2019-10-12
feat(crd): Unified CRD

ncknt:
- Update to latest operator-sdk and k8s 1.14.1
- Move parsing to SpinnakerConfig directly
- Add "inspect" module
- Use operator-sdk as a tool (remove dependency on operator-sdk installed)
- Remove change detector and watcher for config
- Use FreeForm for dynamic parts of the config
- Set operator target
- Ability to override unknown parts of the config


pkg/controller/spinnakerservice/spinnakerservice_controller.go
```diff
@@ -35,8 +32,8 @@ func Add(mgr manager.Manager) error {
 }
 
 type deployer interface {
-	IsSpinnakerUpToDate(ctx context.Context, svc spinnakerv1alpha1.SpinnakerServiceInterface, config runtime.Object, hc *halconfig.SpinnakerConfig) (bool, error)
-	Deploy(ctx context.Context, svc spinnakerv1alpha1.SpinnakerServiceInterface, scheme *runtime.Scheme, config runtime.Object, hc *halconfig.SpinnakerConfig) error
+	IsSpinnakerUpToDate(ctx context.Context, svc spinnakerv1alpha1.SpinnakerServiceInterface) (bool, error)
```
german-muzquiz: This is awesome

pkg/controller/spinnakerservice/status.go
```diff
@@ -3,62 +3,69 @@ package spinnakerservice
...
-	r, err := labels.NewRequirement("app.kubernetes.io/managed-by", selection.Equals, []string{"halyard"})
-	if err != nil {
-		return err
-	}
-	sel := labels.NewSelector()
-	sel.Add(*r)
 	// Get current deployment owned by the service
 	list := &appsv1.DeploymentList{}
-	err = s.client.List(context.TODO(), &client.ListOptions{LabelSelector: sel, Namespace: instance.GetNamespace()}, list)
+	err := s.client.List(context.TODO(), list, client.InNamespace(instance.GetNamespace()), client.MatchingLabels{"app.kubernetes.io/managed-by": "halyard"})
```
german-muzquiz: Maybe this `app.kubernetes.io/managed-by` annotation would be better off having managed-by `spinnaker-operator`?

pkg/controller/spinnakerservice/status.go
```diff
@@ -3,62 +3,69 @@ package spinnakerservice
...
+				Replicas:            it.Status.Replicas,
+				UpdatedReplicas:     it.Status.UpdatedReplicas,
+				ReadyReplicas:       it.Status.ReadyReplicas,
+				AvailableReplicas:   it.Status.AvailableReplicas,
+				UnavailableReplicas: it.Status.UnavailableReplicas,
+				LastUpdateTime:      it.ObjectMeta.CreationTimestamp,
+			}
+			// Spinnaker not ready if any of the services has zero pods
+			if it.Status.ReadyReplicas == 0 {
+				status.Status = Unavailable
```
german-muzquiz: I'm curious what's the `ReadyReplicas` value when a deployment is starting. Let's say clouddriver pod is starting and doing a `kubectl get pods` shows that it has `0/1` pods ready. Doesn't this mean that Spinnaker is in status `Updating` instead of status `Unavailable`?
I just want to make sure that status `Unavailable` is only used when one or more spinnaker service deployments are down, but not while a new pod is being started while the old version is still running.

pkg/controller/spinnakervalidating/gvk.go
```diff
@@ -2,74 +2,40 @@ package spinnakervalidating
...
+func (v *spinnakerValidatingController) getSpinnakerServices() ([]v1alpha2.SpinnakerServiceInterface, error) {
 	list := SpinnakerServiceBuilder.NewList()
-	var opts *client.ListOptions
+	var opts client.ListOption
 	ns, _ := k8sutil.GetWatchNamespace()
 	if ns == "" {
-		opts = &client.ListOptions{}
-	} else {
-		opts = &client.ListOptions{Namespace: ns}
+		opts = client.InNamespace(ns)
```
german-muzquiz: Shouldn't this be 
```
if ns != "" {
    opts = client.InNamespace(ns)
}
```
Seems that this will list in namespace ""

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -23,66 +18,35 @@ import (
...
-			},
-		},
-	})
-
-	if err != nil {
-		return err
-	}
-	return as.Register(validatingWebhook)
+	hookServer := m.GetWebhookServer()
+	hookServer.Register("/validate-v1alpha1-spinnakerservice", &webhook.Admission{Handler: &spinnakerValidatingController{}})
```
german-muzquiz: Where are you specifying for which `kinds` is the webhook invoked? Also, how is the kubernetes `Service` created? It was created by the webhook BootstrapOptions.

pkg/deployer/changedetector/config.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+type configChangeDetectorGenerator struct {
+}
+
+func (g *configChangeDetectorGenerator) NewChangeDetector(client client.Client, log logr.Logger) (ChangeDetector, error) {
+	return &configChangeDetector{client: client, log: log}, nil
+}
+
+// IsSpinnakerUpToDate returns true if there is a x509 configuration with a matching service
```
german-muzquiz: Doc string needs to be updated

pkg/deployer/transformer/target.go
```diff
@@ -32,5 +30,9 @@ func (g *targetTransformerGenerator) GetName() string {
 
 // TransformConfig is a nop
 func (t *targetTransformer) TransformConfig(ctx context.Context) error {
-	return t.hc.SetHalConfigProp("deploymentEnvironment.location", t.svc.GetNamespace())
+	err := t.svc.GetSpinnakerConfig().SetHalConfigProp("deploymentEnvironment.location", t.svc.GetNamespace())
+	if err != nil {
+		return err
+	}
+	return t.svc.GetSpinnakerConfig().SetHalConfigProp("deploymentEnvironment.type", "Operator")
```
german-muzquiz: Are we supporting now this new deployment type? Are there any side effects or other places (like Halyard) that need to be updated?

tools.go
```diff
@@ -11,5 +11,5 @@ import (
 	_ "k8s.io/code-generator/cmd/lister-gen"
 	_ "k8s.io/gengo/args"
 	_ "k8s.io/kube-openapi/cmd/openapi-gen"
-	_ "sigs.k8s.io/controller-tools/pkg/crd/generator"
+	//_ "sigs.k8s.io/controller-tools/pkg/crd/generator"
```
german-muzquiz: I think this line can be deleted

deploy/crds/spinnaker.io_spinnakerservices_crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1alpha2
+apiVersion: apiextensions.k8s.io/v1beta1
```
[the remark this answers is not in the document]
ncknt: No this is the CRD. I made the mistake earlier too. v1beta1 is for `CustomResourceDefinition`, v1alpha2 is referenced further down.

pkg/deployer/transformer/target.go
ncknt: We already had the `Operator` deployment type. We now use it explicitly in Halyard. It's basically `Distributed` but without an account check  (you can deploy without having an account referenced)

pkg/deployer/changedetector/config.go
ncknt: Updated

pkg/controller/spinnakervalidating/gvk.go
ncknt: Good catch, thanks!

pkg/controller/spinnakerservice/status.go
ncknt: Correct. The next statement should make it `unavailable` though since you'll have at least one updated replica that wouldn't be ready. So overall it would be `updating`.

PR #31, opened 2019-10-28
feat(example): refactor and added comments to kustomize

kevinawoo:


deploy/spinnaker/kustomize/blah.yml
```diff
@@ -0,0 +1,10 @@
+apiVersion: spinnaker.io/v1alpha2
```
ncknt: Can we use a different file name than `blah`?

PR #36, opened 2019-10-30
feat(crd): Account definitions and CRD

ncknt:


pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
```diff
@@ -0,0 +1,85 @@
...
+	GetEnv() map[string]interface{}
+	GetSettings() map[string]interface{}
+	GetPermissions() *AccountPermissions
+}
+
+// SpinnakerAccountSpec defines the desired state of SpinnakerAccount
+// +k8s:openapi-gen=true
+type SpinnakerAccountSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
```
german-muzquiz: This should be `make k8s openapi` right?

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
```diff
@@ -0,0 +1,85 @@
...
+	Auth        FreeForm           `json:"auth,omitEmpty"`
+	Env         FreeForm           `json:"env,omitEmpty'`
+	Settings    FreeForm           `json:"settings,omitEmpty"`
+}
+
+// SpinnakerAccountStatus defines the observed state of SpinnakerAccount
+// +k8s:openapi-gen=true
+type SpinnakerAccountStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
```
german-muzquiz: Also here for the `make` goal instead of operator-sdk binary

pkg/controller/spinnakeraccount/spinnakeraccount_controller.go
```diff
@@ -0,0 +1,148 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("spinnakeraccount-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource SpinnakerService
+	err = c.Watch(&source.Kind{Type: &v1alpha2.SpinnakerAccount{}}, &handler.EnqueueRequestForObject{})
```
german-muzquiz: Do we need to watch for `SpinnakerService` in case the acounts enabled/dynamic flag is changed?

pkg/validate/providers.go
```diff
@@ -0,0 +1,89 @@
+package validate
```
german-muzquiz: Suggestion: rename this file from `providers.go` to `accounts.go`, since there are account types that are not a provider (jenkins, slack, etc.)

pkg/validate/providers_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"github.com/armory/spinnaker-operator/pkg/accounts/kubernetes"
+	"github.com/armory/spinnaker-operator/pkg/apis/spinnaker/v1alpha2"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestGetAccountsFromConfig(t *testing.T) {
+	spinsvc := &v1alpha2.SpinnakerService{
```
german-muzquiz: There are already helper functions in `test` module to load test manifest yaml files, so you don't need to code the whole `SpinnakerService` structure in each test case:
```
spinsvc := test.ManifestToSpinService(manifestFilePath, t)
```
Also, it makes it easier to replicate bugs when you have a sample manifest file to include quickly in a new test case.
We can follow the same approach for tests related with account CRD's.

pkg/validate/validate.go
```diff
@@ -5,39 +5,21 @@ import (
...
-// allValidatorsInSequence is used to register all SpinnakerValidator in the right execution order
-var allValidatorsInSequence []SpinnakerValidator
-
-// accountValidators keeps track of all existing AccountValidator
-var accountValidators []AccountValidator
-
-func init() {
-	accountValidators = []AccountValidator{&kubernetesAccountValidator{}}
-	allValidatorsInSequence = []SpinnakerValidator{
-		&singleNamespaceValidator{},
```
german-muzquiz: I think we need to keep a clean way of seeing the full sequence for which validators will run. With thew new approach, it's not possible to tell easily what will be run when without looking at the implementation. Also for adding new validators, it would be easier to insert in the desired sequence.

deploy/crds/spinnaker.io_spinnakerservices_crd.yaml
```diff
@@ -59,6 +59,15 @@ spec:
         spec:
           description: SpinnakerServiceSpec defines the desired state of SpinnakerService
           properties:
+            accounts:
+              properties:
+                dynamic:
```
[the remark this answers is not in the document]
ncknt: enabled -> use CRDs. dynamic -> inject at runtime

pkg/accounts/account/account.go
```diff
@@ -0,0 +1,34 @@
...
+	// Key under which accounts are stored
+	GetAccountsKey() string
+}
+
+type Account interface {
+	GetName() string
+	GetType() v1alpha2.AccountType
+	NewValidator() AccountValidator
+	ToSpinnakerSettings() (map[string]interface{}, error)
+	GetEnv() interface{}
```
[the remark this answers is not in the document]
ncknt: All accounts are defined this way. It contains the "environment" of the account. Namespaces, etc. Settings was already taken :)

pkg/accounts/account/account.go
```diff
@@ -0,0 +1,34 @@
...
+	GetAuth() interface{}
+	GetSettings() *v1alpha2.FreeForm
+	GetHash() (string, error)
+}
+
+type AccountValidator interface {
+	Validate(v1alpha2.SpinnakerServiceInterface, client.Client, context.Context) error
```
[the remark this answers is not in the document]
ncknt: That's on purpose to avoid extra dependencies. `ValidationResult` won't be used in the CRD world.

pkg/accounts/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+	// Cluster to use in the kubeconfig file
+	Cluster        string `json:"cluster,omitempty"`
+	ServiceAccount bool   `json:"serviceAccount,omitempty"`
+	// Reference to a kubeconfig file
+	KubeconfigFile      string   `json:"kubeconfigFile,omitempty"`
+	OAuthServiceAccount string   `json:"oAuthServiceAccount,omitempty"`
+	OAuthScopes         []string `json:"oAuthScopes,omitempty"`
+}
+
+type Env struct {
```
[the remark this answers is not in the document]
ncknt: That's the plan. `Env` won't necessarily be used.

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
ncknt: That's left over from the code generation.

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
ncknt: Same as above

pkg/controller/spinnakeraccount/spinnakeraccount_controller.go
ncknt: I don't think so. If `SpinnakerService` changes, it will be reconciled and get the CRD accounts on its own.

pkg/validate/providers.go
ncknt: So many `accounts.go` but makes sense.

pkg/validate/providers_test.go
ncknt: I see the value for larger manifests but I find that much less readable. It's also hard to see where the file is used. We should probably add as a simple parse with a given string. What do you think?

pkg/validate/validate.go
ncknt: It's basically singleNamespace then everything else. I'd rather keep the simplicity and if we do need more logic, add a more formal validation workflow. As you can see, we cannot define things statically now as validators are determined at runtime.

deploy/crds/spinnaker.io_spinnakerservices_crd.yaml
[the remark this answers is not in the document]
german-muzquiz: If both these settings only apply to accounts defined in CRD, maybe we can rename this field to `crdAccounts`, otherwise it could cause confusion to have `accounts.enabled: false` but still have working all accounts defined in main config file/profile files.

pkg/validate/providers_test.go
german-muzquiz: Parse with a given string is ok for smaller structures.

deploy/crds/spinnaker.io_spinnakerservices_crd.yaml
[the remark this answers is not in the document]
ncknt: I'd like to avoid calling anything `crd` as it's not even the right term. `SpinnakerAccounts` may be better?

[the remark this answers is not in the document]
german-muzquiz: Better!

PR #39, opened 2019-10-31
feat(validations): Spinnaker version validation

german-muzquiz:


pkg/halyard/halyard.go
```diff
@@ -17,43 +18,48 @@ import (
 
 // Service is the Halyard implementation of the ManifestGenerator
 type Service struct {
-	url string
+	halyardBaseUrl string
```
ncknt: Can we keep `url` it's more Goish?

PR #43, opened 2019-10-31
fix(build): Fix release version on Jenkins

ncknt:


Jenkinsfile
```diff
@@ -1,3 +1,21 @@
+@NonCPS
```
kevinawoo: 🚔

PR #45, opened 2019-10-31
feat(validation): multiple val webhooks

ncknt:


german-muzquiz:
Just wanting to be sure, is the old validations webhook still working?

ncknt:
Yes

PR #48, opened 2019-11-01
feat(deploy): allow multi-layer deploy actions on spinsvc

ncknt:


pkg/controller/spinnakerservice/spinnakerservice_controller.go
```diff
@@ -21,30 +24,32 @@ import (
...
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
 	return add(mgr, newReconciler(mgr))
 }
 
-type deployer interface {
-	IsSpinnakerUpToDate(ctx context.Context, svc spinnakerv1alpha2.SpinnakerServiceInterface) (bool, error)
-	Deploy(ctx context.Context, svc spinnakerv1alpha2.SpinnakerServiceInterface, scheme *runtime.Scheme) error
-}
+type deployerGenerator func(m deploy.ManifestGenerator, mgr manager.Manager, clientset *kubernetes.Clientset, logger logr.Logger) deploy.Deployer
```
german-muzquiz: Could this be moved to `deploy` module, so that both deployer and its generator live in the same place?

pkg/deploy/deploy.go
```diff
@@ -0,0 +1,19 @@
+package deploy
+
+import (
+	"context"
+	spinnakerv1alpha2 "github.com/armory/spinnaker-operator/pkg/apis/spinnaker/v1alpha2"
+	"github.com/armory/spinnaker-operator/pkg/generated"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type ManifestGenerator interface {
```
german-muzquiz: This will get injected into all deployers, but the `Generate` method seems too coupled with just deploying Spinnaker (returns a `SpinnakerGeneratedConfig`). What if this interface is just `Halyard` so that all deployers can call any Halyard method if needed? Halyard service instance is also passed to validators using their `Options` object.

pkg/controller/spinnakerservice/spinnakerservice_controller.go
ncknt: We don't need it in the other module. I'd argue the generator per se is only used here. in `deploy/spindeploy` we only have a regular `NewXXX` function

pkg/deploy/deploy.go
ncknt: Yeah that was the intent (make Halyard(ish) service available to all) We could enhance the interface as needed. I'd rather not use `halyard` directly to not couple these two (we have too many of these dependencies already!)

PR #49, opened 2019-11-02
feat(account): hook account validation

ncknt:
- Start Kubernetes account validation
- Support secret file decryption and validate file exists

### DIFF
--- a/pkg/accounts/kubernetes/validator.go
+++ b/pkg/accounts/kubernetes/validator.go
@@ -2,7 +2,12 @@ package kubernetes
 
 import (
 	"context"
+	"fmt"
 	"github.com/armory/spinnaker-operator/pkg/apis/spinnaker/v1alpha2"
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/authorization/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/accounts/kubernetes/validator.go
```diff
@@ -2,14 +2,96 @@ package kubernetes
...
+	// Checks are by namespace if namespaces or omitNamespaces (get them before)
+	// Get list of api resources (see getResources below)
+	// If CRDs are listed, try them explicitly
+	sars := clientset.AuthorizationV1().SelfSubjectAccessReviews()
+	sar := &v1.SelfSubjectAccessReview{
+		Spec: v1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &v1.ResourceAttributes{
+				//Namespace:   "",
+				Verb:    "*",
+				Group:   "*",
```
german-muzquiz: I'm not sure if this is taken into account by the "TODO" note above, but the code that actually runs a `kubeconfig auth can-i --list` uses a `SelfSubjectRulesReview` that the only parameter that it needs is the namespace: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/auth/cani.go#L219 . Sample output:
```
Resources                                       Non-Resource URLs        Resource Names     Verbs
*.*                                             []                       []                 [*]
                                                [*]                      []                 [*]
selfsubjectaccessreviews.authorization.k8s.io   []                       []                 [create]
selfsubjectrulesreviews.authorization.k8s.io    []                       []                 [create]
                                                [/api/*]                 []                 [get]
                                                [/api]                   []                 [get]
                                                [/apis/*]                []                 [get]
                                                [/apis]                  []                 [get]
                                                [/healthz]               []                 [get]
                                                [/openapi/*]             []                 [get]
                                                [/openapi]               []                 [get]
                                                [/swagger-2.0.0.pb-v1]   []                 [get]
                                                [/swagger.json]          []                 [get]
                                                [/swaggerapi/*]          []                 [get]
                                                [/swaggerapi]            []                 [get]
                                                [/version/]              []                 [get]
                                                [/version]               []                 [get]
podsecuritypolicies.policy                      []                       [eks.privileged]   [use]
```
Maybe we can use it instead of specifying verb, group, etc. to issue as less requests as possible.

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
```diff
@@ -24,13 +24,6 @@ const (
 
 type AccountPermissions map[Authorization][]string
 
-type SpinnakerAccountInterface interface {
-	GetAuth() map[string]interface{}
-	GetEnv() map[string]interface{}
-	GetSettings() map[string]interface{}
-	GetPermissions() *AccountPermissions
-}
```
german-muzquiz: Why was this interface removed?

ncknt: Not used

pkg/accounts/kubernetes/validator.go
ncknt: Yeah it's part of the kubernetes validation task. We need a more fine grained than a blanked `can-i` IMO. At the very least by namespaces when they're defined. Also taking into accounts kind, omitKind, and omitNamespace. It's a skeleton of the validation.

PR #52, opened 2019-11-05
fix(crd/validation): config hash as map

ncknt:
- merge master

### DIFF
--- a/pkg/deploy/spindeploy/changedetector/config.go
+++ b/pkg/deploy/spindeploy/changedetector/config.go
@@ -7,6 +7,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const SpinnakerConfigHashKey = "Config"
+
 type configChangeDetector struct {
 	client client.Client
 	log    logr.Logger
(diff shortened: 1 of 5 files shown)


pkg/deploy/spindeploy/changedetector/config.go
```diff
@@ -25,7 +27,15 @@ func (ch *configChangeDetector) IsSpinnakerUpToDate(ctx context.Context, spinSvc
 	if err != nil {
 		return false, err
 	}
-	eh := spinSvc.GetStatus().LastConfigHash
-	spinSvc.GetStatus().LastConfigHash = h
+	st := spinSvc.GetStatus()
```
german-muzquiz: I see that this function is doing a modification of the status as well as checking if there were any changes.
What happens if the config was changed, this function sets the new hash, but then there's an error applying the changes, would the lastDeployedHashes be out of sync with what is actually deployed? Would it be better to move updating the last deployed hash after the deployment is applied?

ncknt: Then the status wouldn't be saved. We're working on a clone here (normally).

ncknt: To clarify the status should be updated as part of `commitConfigToStatus` after all manifests are saved.

PR #53, opened 2019-11-06
feat(validation): use validation settings

ncknt:
- Moved hash to a map of hashes where we'll store the result of account validation
- Added NeedsValidation to implement `validation.frequencySeconds`
- use `fatal` on account validator to mute validation errors


pkg/apis/spinnaker/v1alpha2/spinnakerservice_types.go
```diff
@@ -124,42 +151,47 @@ type SpinnakerServiceStatus struct {
...
-// +kubebuilder:resource:path=spinnakerservices,shortName=spinsvc
-type SpinnakerService struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
-
-	Spec   SpinnakerServiceSpec   `json:"spec,omitempty"`
-	Status SpinnakerServiceStatus `json:"status,omitempty"`
+// GetUpdateHash updates the hash at key `key` and returns the prior copy if one existed
+// LastDeployed should then contain the hash and the time if updateTime is true or if there was no hash
+func (s *SpinnakerServiceStatus) UpdateHashIfNotExist(key, hash string, t time.Time, updateTime bool) HashStatus {
```
german-muzquiz: I think we could move these functions (UpdateHashIfNotExist, GetValidationSettings and NeedsValidation) to some other file, to keep this "types" file only concerned in defining types for the CRD.

pkg/validate/accounts.go
```diff
@@ -2,24 +2,40 @@ package validate
...
+		now := time.Now()
 		for _, a := range as {
-			validators = append(validators, &accountValidator{v: a.NewValidator()})
+			h, err := a.GetHash()
+			if err != nil {
+				return nil, err
+			}
+			hc := status.UpdateHashIfNotExist(getValidationHashKey(a), h, now, false)
+			// If accounts were never validated or if the validation is too old less than x ago
+			if hc.Hash != h || !v.NeedsValidation(hc.LastUpdatedAt) {
```
german-muzquiz: Shouldn't `!v.NeedsValidation(hc.LastUpdatedAt)` be NOT negated here?

ncknt: Oops thanks!

PR #54, opened 2019-11-07
doc(uninstall): Add uninstall instructions keeping spinnaker

german-muzquiz:


README.md
```diff
@@ -114,3 +114,65 @@ spinnakerservice.spinnaker.io "spinnaker" deleted
...
+  <<CONTENT HERE>> 
+```
+2. For each entry in `spec.spinnakerConfig.profiles`, copy it to its own file inside a `profiles` folder with a `<entry-name>-local.yml` name.
+3. For each entry in `spec.spinnakerConfig.service-settings`, copy it to its own file inside a `service-settings` folder with a `<entry-name>.yml` name.
+4. For each entry in `spec.spinnakerConfig.files`, copy it to its own file inside a directory structure following the name of the entry with double underscores (__) replaced by a path separator. Example: an entry named `profiles__rosco__packer__example-packer-config.json` would produce the file `profiles/rosco/packer/example-packer-config.json`.
+
+At the end, you would have the following directory tree:
+```
+config
+profiles/
```
ncknt: Missing `default/`

README.md
```diff
@@ -114,3 +114,65 @@ spinnakerservice.spinnaker.io "spinnaker" deleted
...
+service-settings/
+```
+
+After that, you can put these files in your Halyard home directory and deploy Spinnaker running `hal deploy apply`.
+
+Finally you can delete the operator and their CRD's from the Kubernetes cluster
```
ncknt: `CRD's` -> `CRDs`

README.md
```diff
@@ -114,3 +114,65 @@ spinnakerservice.spinnaker.io "spinnaker" deleted
...
+
+```bash
+$ kubectl delete -n <namespace> -f deploy/operator/<installation type>
+$ kubectl delete -f deploy/crds/
+```
+
+### Removing operator ownership from Spinnaker resources
+
+Removing ownership can de done executing a `kubectl edit <resource type> <resource name>`, and deleting the entry in block `metadata.ownerReferences` that includes `SpinnakerService` kind. Example:
```
ncknt: Could we do it with a script? (Get all secrets and deployments) and remove the `ownerReferences`. Also editing the deployment should redeploy.

README.md
ncknt: Or and period.

README.md
german-muzquiz: `service` don't have the same labels as `deployment` for getting them in an automated way :/
```
apiVersion: v1                                                                                                                                                                                     
kind: Service
metadata:
  creationTimestamp: "2019-11-06T21:21:42Z"
  labels:
    app: spin
    cluster: spin-deck
```
We can do a select by `app: spin` instead, not ideal but we can assume this app name to be constant.
As for actually deleting the ownerReference, is there a way to do that in shell? My only guess is outputting the manifest as json/yaml, parse it in some way and apply it.

ncknt: By script, I meant just a command with `kubectl`, `jsonpath` and `kubectl apply` back. I'll approve, feel free to keep it for later.

PR #55, opened 2019-11-09
feat(accounts): account definition, CRD, validation

ncknt:
- start of Kubernetes account validation/parsing


pkg/accounts/kubernetes/validator.go
```diff
@@ -4,36 +4,166 @@ import (
...
+	}
+	if settings.Cluster != "" {
+		if cluster, ok := config.Clusters[settings.Cluster]; ok {
+			overrides.ClusterInfo = *cluster
+		}
+	}
+	if len(settings.OAuthScopes) > 0 {
+		overrides.AuthInfo = clientcmdapi.AuthInfo{
+			AuthProvider: &clientcmdapi.AuthProviderConfig{
+				Name: "gcp",
```
german-muzquiz: Is this specific to GCP?

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
```diff
@@ -27,19 +28,40 @@ type AccountPermissions map[Authorization][]string
...
+	Type    AccountType `json:"type"`
+	// +optional
+	Validation ValidationSetting `json:"validation"`
+	// +optional
 	Permissions AccountPermissions `json:"permissions"`
-	Auth        FreeForm           `json:"auth,omitempty"`
-	Env         FreeForm           `json:"env,omitempty"`
-	Settings    FreeForm           `json:"settings,omitempty"`
+	// +optional
+	Kubernetes *KubernetesAuth `json:"kubernetes,omitempty"`
```
german-muzquiz: It seems strange to have provider specific information here that only works for Kubernetes, when there is a generic `Settings` field and a `Type` field. Why not have then all Kubernetes account settings here and not just the auth?, and structs for AWS, GCP, Jenkins and so on...

pkg/secrets/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,59 @@
...
+	return sec, nil
+}
+
+func (sd *SecretDecrypter) fetchSecret(secret Secret) (string, error) {
+	sec := &v1.Secret{}
+	if err := sd.client.Get(context.TODO(), client.ObjectKey{Name: secret.name, Namespace: sd.namespace}, sec); err != nil {
+		return "", err
+	}
+	if d, ok := sec.Data[secret.key]; ok {
+		return string(d), nil
```
german-muzquiz: Is this actually decoding the secret? Seems that it just returns the base64 string.

pkg/secrets/nop.go
```diff
@@ -0,0 +1,13 @@
+package secrets
+
+type nopDecrypter struct {
```
german-muzquiz: What's the purpose of this?

pkg/secrets/secrets.go
```diff
@@ -33,37 +36,56 @@ func DecodeAsFile(ctx context.Context, val string) (string, error) {
...
+func NewDecrypter(c client.Client, namespace string, encryptedSecret string) (secrets.Decrypter, error) {
+	engine, params := secrets.ParseTokens(encryptedSecret)
+	switch engine {
+	case "s3":
+		return secrets.NewS3Decrypter(params), nil
+	case "vault":
+		return secrets.NewVaultDecrypter(params), nil
+	case "k8s":
+		return kubernetes.NewSecretDecrypter(c, namespace, params), nil
+	case "noop":
```
german-muzquiz: What's a `noop` secret?

pkg/accounts/kubernetes/validator.go
ncknt: Yes. It uses the gcp kubectl plugin. This needs to be validated as part of the Kubernetes validation story.

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
ncknt: That's to have a more typed spec. That's also how we inline the kubeconfig. Other providers won't have as many options but I'd still like to make them typed instead of having a hodgepodge of settings.

pkg/secrets/kubernetes/kubernetes.go
ncknt: It does return the original content. The secret is decoded by the client - at least in my tests. Give it a try.

pkg/secrets/nop.go
ncknt: testing via a secret engine that just returns the content.

pkg/secrets/secrets.go
ncknt: noop = nop above. No operation decrypted.

german-muzquiz: I'm still not sure how this works in real life or why is needed, is it just to support some test cases? there could be another way to test without creating a dummy secrets engine.

pkg/apis/spinnaker/v1alpha2/spinnakeraccount_types.go
german-muzquiz: Maybe I'm not seeing the full picture here and this is going to change, but I think it's confusing to tell people that all account configuration settings belong to a generic free format section shared between all accounts, while just one setting, of one specific account type, has its own special field handled as a special case. 

We could have all kubernetes settings included in the kubernetes field as well, a free format section for everything and a kubeconfig or auth section for the kubeconfig setting:
```
spec:
  type: kubernetes
  kubernetes:
    auth:
      kubeconfigfile: XXX
    settings: <free format>
```
In this way we can add more types accounts as we need them by having just one level or two of strong typing.

ncknt: Ah I see. To be clear we're talking of:
```
spec:
  type: kubernetes
  permissions: {} 
  kubernetes:
      kubeconfigfile: XXX
    settings: <free format>
```
vs
```
spec:
  type: kubernetes
  permissions: {} 
  kubernetes:
      kubeconfigfile: XXX
  settings: <free format>
```

We have other shared top level fields so I'm not sure why settings is special but maybe I've been looking at it for too long. Let's get some feedback.

pkg/secrets/secrets.go
ncknt: I use it for testing and don't intent to publish it. The only use case I'd see is someone inlining the kubeconfig as a string and not wanting to use a Kubernetes secret.

PR #56, opened 2019-11-18
fix(validations/hal): Externalize non implemented validations to halyard

ncknt:
- Expand validation settings
- Remove verbose change detector


pkg/halyard/validate.go
```diff
@@ -0,0 +1,85 @@
...
+func parseValidationResponse(d []byte, logger logr.Logger) error {
+	resp := make(validationResponse, 0)
+	if err := json.Unmarshal(d, &resp); err != nil {
+		return errors.Wrap(err, "unable to read external validation response")
+	}
+	msgs := make([]string, 0)
+	for _, v := range resp {
+		if v.Severity == "WARNING" {
+			logger.Info(fmt.Sprintf("WARN: %s at %s", v.Message, v.Location))
+		} else {
```
german-muzquiz: The code reads like "throw an error if severity is anything but warnings". It would be better to explicitly error on severities higher than warnings, or not error on severity <= warninng.

pkg/halyard/validators.go
```diff
@@ -0,0 +1,173 @@
+package halyard
+
+import "github.com/armory/spinnaker-operator/pkg/apis/spinnaker/v1alpha2"
+
+// Validators we're going always going to skip
```
german-muzquiz: Can you document that all validator names should exactly match the class names in Halyard extending `Validator` class?
Also this seems like a fragile solution, ideally it should be possible to just run all validators without actually naming them, otherwise we would need to constantly update this file every time someone adds/deletes or renames a validator class in Halyard, this could get out of sync pretty quickly.

ncknt: Here we just list the validations that you do NOT want to run (those the operator implements itself).

german-muzquiz: I know that, but it should be commented in the code, there's no way someone without the context will know that validator names should match exactly Java class names in Halyard

ncknt: I can add a comment that this should match the Halyard class name. I was confused by the other part of the comment. Yes if validation class names change, they'd not be skipped. But I don't recall that happening and the operator is tied to Halyard for now anyway. When we implement all validations in the operator, we won't be calling out to Halyard anymore.

pkg/halyard/validate.go
ncknt: Changed!

pkg/halyard/validators.go
ncknt: Added comment

PR #59, opened 2019-11-21
feat(secrets/k8s): upstream secrets changes, sanitization, file cache

ncknt:
- Proper registration of the Kubernetes secret engine


pkg/halyard/halyard.go
```diff
@@ -106,6 +91,14 @@ func (s *Service) buildGenManifestsRequest(ctx context.Context, spinConfig *v1al
 	return req, nil
 }
 
+func (s *Service) addObjectToRequest(ctx context.Context, writer *multipart.Writer, param string, object interface{}) error {
```
german-muzquiz: Looks like `ctx` parameter is not used

pkg/secrets/kubernetes.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (k *KubernetesDecrypter) parse(params string) error {
+	tokens := strings.Split(params, "!")
+	for _, element := range tokens {
+		kv := strings.Split(element, ":")
+		if len(kv) == 2 {
+			switch kv[0] {
+			case "n":
+				k.name = kv[1]
```
german-muzquiz: Would it be useful to include secret namespace as an optional param (defaulting to what you have now)?

pkg/secrets/secrets.go
```diff
@@ -2,90 +2,68 @@ package secrets
...
 	"os"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"strings"
 )
 
-// Decode decodes a potential value into a secret
-func Decode(ctx context.Context, val string) (string, error) {
-	return decode(decryptFunc, ctx, val)
+func init() {
+	secrets.Engines["kubernetes"] = NewSecretDecrypter
```
german-muzquiz: Function name could be changed to something indicating it's a kubernetes specific decrypter, `NewKubeSecretDecrypter`?

pkg/secrets/kubernetes.go
ncknt: It's by design, we don't want to mount secrets from a different namespace.

pkg/secrets/secrets.go
ncknt: Oh yeah, it used to be in a different module. Will rename

pkg/halyard/halyard.go
ncknt: It used to be :) Removed

PR #60, opened 2019-11-21
feat(validation/k8s): settings check, namespace access check 

ncknt:
- Add inspect.GetStringArray()


pkg/accounts/kubernetes/validator.go
```diff
@@ -193,61 +196,26 @@ func (k *kubernetesAccountValidator) validateAccess(clientConfig clientcmd.Clien
 	if err != nil {
 		return err
 	}
+	// Get namespaces. The test is analogous to what is done in Halyard
+	// We want to keep it short so any improvement should remain short (e.g. not a request per namespace)
+	_, err = clientset.CoreV1().Namespaces().List(v13.ListOptions{})
```
german-muzquiz: I think that when you supply an explicit list of namespaces in the account configuration, then clouddriver doesn't even try the list namespaces call (because it already knows them), so maybe we can replace this check in that case and verify access to the cluster with a get xxx in only one of the namespaces supplied. The problem with the list namespaces call is that I think it will fail with service accounts that only have access to one explicit target namespace.

ncknt: Clouddriver doesn't but Halyard does perform that validation. I think get namespace will fail only if your service account is not authorized to list the namespaces not if it not authorized to perform other actions in that namespace (if that makes sense). That's at least how I take this: https://github.com/kubernetes/community/issues/1486

Also what would xxx be?

ncknt: I'm second guessing that now. That said, this is the validation that Halyard currently does and AFAIK it works.

german-muzquiz: Ah, if that's what Halyard is doing today I think we can reuse it

PR #63, opened 2019-11-22
fix(secrets/k8s): sanitize secrets before validation calls

ncknt:
This resolves any referenced secret before the config is sent to Halyard for validation. Passwords/tokens are directly replaced with their resolved value. File references (only supported via `encryptedFile`) are changed to `secrets/<the local temp file name>`. That secret file is then added to the config sent to Halyard under the `secrets` key.

- sanitize: interfaces and use relative secret file path

### DIFF
--- a/pkg/inspect/parse.go
+++ b/pkg/inspect/parse.go
@@ -164,11 +164,7 @@ func sanitizeSecretsReflect(ctx context.Context, v reflect.Value, stringHandler 
 		}
 		return nmv, nil
 	case reflect.Interface:
-		iv, err := sanitizeSecretsReflect(ctx, v.Elem(), stringHandler)
-		if err != nil {
-			return v, err
-		}
-		return iv, nil
+		return sanitizeSecretsReflect(ctx, v.Elem(), stringHandler)
 	}
 	return v, nil
 }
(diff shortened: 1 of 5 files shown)


pkg/inspect/parse.go
```diff
@@ -81,20 +81,27 @@ func toSpecificArray(array reflect.Value, target reflect.Type) (reflect.Value, e
...
-
 // SanitizeSecrets visits all nodes and returns copies of the struct with secrets that are not passthrough
-// replaced
-func SanitizeSecrets(ctx context.Context, i interface{}) (interface{}, error) {
-	t, err := sanitizeSecretsReflect(ctx, reflect.ValueOf(i), secretHandler)
+// replaced. Regular token are replaced, files are replaced with the file name prefixed by relativeSecretPath
+// e.g.
+// a.b: encrypted:xxx -> a.b: <decrypted value>
+// a.b: encryptedFile:xxx -> a.b: relativeSecretPath + temp file name used
+func SanitizeSecrets(ctx context.Context, relativeSecretPath string, i interface{}) (interface{}, error) {
```
german-muzquiz: How is this different than decrypting all secrets in a struct? Maybe a better name would be `DecryptSecrets` instead of `SanitizeSecrets`?

ncknt: It has changed in the next PR.

PR #64, opened 2019-11-22
feat(kubernetes/sa): Validate service account

german-muzquiz:


pkg/accounts/kubernetes/validator.go
```diff
@@ -32,21 +36,45 @@ type kubernetesAccountValidator struct {
...
 	return k.validateAccess(config)
 }
 
-func (k *kubernetesAccountValidator) makeClient(ctx context.Context, spinSvc v1alpha2.SpinnakerServiceInterface) (clientcmd.ClientConfig, error) {
+func (k *kubernetesAccountValidator) ensureSpinSvc(spinSvc v1alpha2.SpinnakerServiceInterface, c client.Client) (v1alpha2.SpinnakerServiceInterface, error) {
+	if spinSvc != nil {
+		return spinSvc, nil
+	}
+	i := SpinnakerServiceBuilder.NewList()
+	list, err := util.GetSpinnakerServices(i, k.account.CrdNamespace, c)
```
ncknt: use `spinSvc.getNamespace()`. An account must be in the same namespace as the spinnaker service. `CrdNamespace` is not necessary.

pkg/accounts/kubernetes/kubernetes.go
```diff
@@ -63,10 +66,12 @@ type CustomKubernetesResource struct {
 
 type Account struct {
 	*account.BaseAccount
-	Name     string `json:"name,omitempty"`
-	Auth     *v1alpha2.KubernetesAuth
-	Env      Env               `json:"env,omitempty"`
-	Settings v1alpha2.FreeForm `json:"settings,omitempty"`
+	Name string `json:"name,omitempty"`
+	// if this account comes from a CRD, this field indicates its namespace
+	CrdNamespace string
```
ncknt: Namespace is available in the context. I'd rather use that as namespace is not an account setting.
```
sec = secrets.FromContextWithError(ctx) -> sec.namespace
```

pkg/accounts/kubernetes/parse.go
```diff
@@ -38,10 +38,35 @@ func (k *AccountType) FromSpinnakerConfig(settings map[string]interface{}) (acco
...
+	}
+	sa, ok := settings["serviceAccount"]
+	if ok {
+		s, sok := sa.(bool)
+		if !sok {
+			return nil, fmt.Errorf("serviceAccount is not a boolean: %s", sa)
+		}
+		return &v1alpha2.KubernetesAuth{UseServiceAccount: s}, nil
+	}
+	return nil, fmt.Errorf("unable to parse account %s: no valid kubeconfig file, kubeconfig content or service account information found", name)
```
ncknt: Users could also be using `kubeconfigFileContent`

pkg/accounts/kubernetes/validator.go
```diff
@@ -32,21 +36,49 @@ type kubernetesAccountValidator struct {
 }
 
 func (k *kubernetesAccountValidator) Validate(spinSvc v1alpha2.SpinnakerServiceInterface, c client.Client, ctx context.Context, log logr.Logger) error {
-	if err := k.validateSettings(ctx, log); err != nil {
+	err := k.validateSettings(ctx, log)
```
ncknt: wondering why this was changed. I generally find better (and more readable but I guess that's subjective) to do:
```
if err := xxx; err != nil {
  return err
}
```
as it doesn't introduce another scope variable - even when another variable is done further down.

pkg/accounts/kubernetes/validator.go
```diff
@@ -32,21 +36,49 @@ type kubernetesAccountValidator struct {
...
+		return nil, err
+	}
+	list, err := util.GetSpinnakerServices(i, sc.Namespace, c)
+	if err != nil {
+		return nil, err
+	}
+	if len(list) == 0 {
+		return nil, nil
+	} else {
+		// there's only one spinnaker service per namespace
```
ncknt: There **should** be one spinnaker if validation was on. I'm fine taking the first one though!

pkg/accounts/kubernetes/validator.go
```diff
@@ -119,26 +155,54 @@ func makeClientFromSettings(ctx context.Context, settings map[string]interface{}
...
 		}
-		return clientcmd.NewDefaultClientConfig(*cfg, makeOverrideFromAuthSettings(cfg, aSettings)), nil
+		return clientcmd.NewDefaultClientConfig(*cfg, makeOverrideFromAuthSettings(cfg, aSettings)).ClientConfig()
 	}
 	return nil, noValidKubeconfigError
 }
 
-func makeClientFromServiceAccount(ctx context.Context, spinSvc v1alpha2.SpinnakerServiceInterface) (clientcmd.ClientConfig, error) {
-	_, err := spinSvc.GetSpinnakerConfig().GetServiceSettingsPropString(ctx, util.ClouddriverName, "kubernetes.serviceAccountName")
+func makeClientFromServiceAccount(ctx context.Context, spinSvc v1alpha2.SpinnakerServiceInterface, c client.Client) (*rest.Config, error) {
```
ncknt: That'd make more sense to look for a SpinnakerService here instead of at the beginning.

pkg/accounts/kubernetes/validator.go
```diff
@@ -32,21 +36,49 @@ type kubernetesAccountValidator struct {
...
 func (k *kubernetesAccountValidator) Validate(spinSvc v1alpha2.SpinnakerServiceInterface, c client.Client, ctx context.Context, log logr.Logger) error {
-	if err := k.validateSettings(ctx, log); err != nil {
+	err := k.validateSettings(ctx, log)
+	if err != nil {
 		return err
 	}
-	if k.account.Auth != nil && k.account.Auth.UseServiceAccount && spinSvc == nil {
-		// don't validate if the spinnaker account needs a k8s service account to run validations, and there's no SpinnakerService yet
-		return nil
+	if k.account.Auth != nil && k.account.Auth.UseServiceAccount {
```
ncknt: Can we do it below in `makeClientFromServiceAccount`? That seems like a more natural place instead of passing `spinsvc` around.

pkg/accounts/kubernetes/validator.go
german-muzquiz: The reason is that further down I'm trying to reassign `spinSvc` the output of a function that returns an error, so I needed to reuse this same `err` to not introduce any new variables.

pkg/accounts/kubernetes/validator.go
german-muzquiz: We still need to pass spinSvc around because it's received as an argument when running validations as part of SpinnakerService itself.

ncknt: Okay. That still seems more logical to do it closer to where we need it.

pkg/accounts/kubernetes/parse.go
```diff
@@ -38,10 +40,49 @@ func (k *AccountType) FromSpinnakerConfig(settings map[string]interface{}) (acco
...
+	if err != nil {
+		return nil, err
+	}
+	a.Auth = auth
 	a.Settings = settings
 	return a, nil
 }
 
+func (k *AccountType) authFromSpinnakerConfig(name string, settings map[string]interface{}) (*v1alpha2.KubernetesAuth, error) {
+	kubeconfigFile, ok := settings["kubeconfigFile"]
```
ncknt: Can we use `inspect.GetObjectPropString(ctx, settings, "kubeconfigFile")`? It'd have the added benefit of supporting secrets.

pkg/accounts/kubernetes/parse.go
```diff
@@ -38,10 +40,49 @@ func (k *AccountType) FromSpinnakerConfig(settings map[string]interface{}) (acco
...
+	}
+	sa, ok := settings["serviceAccount"]
+	if ok {
+		s, sok := sa.(bool)
+		if !sok {
+			return nil, fmt.Errorf("serviceAccount is not a boolean: %s", sa)
+		}
+		return &v1alpha2.KubernetesAuth{UseServiceAccount: s}, nil
+	}
+	kubeContent, ok := settings["kubeconfig"]
```
ncknt: What is `kubeconfig`? Is it for `kubeconfigFileContent`?

german-muzquiz: Misnamed, should be `kubeconfigContents`

PR #74, opened 2019-12-11
docs(operator): copyedit github docs

brian-armory:
I cleaned up some typos and clarity issues. This iteration is cleaner than what's in my docs.armory PR, so I will update that one.

### DIFF
--- a/doc/options.md
+++ b/doc/options.md
@@ -8,38 +8,38 @@ kind: SpinnakerService
 metadata:
   name: spinnaker
 spec:
-  # spec.spinnakerConfig - This section is how to specify configuration spinnaker
+  # spec.spinnakerConfig - This section describes how to specify configurations spinnaker.
   spinnakerConfig:
-    # spec.spinnakerConfig.config - This section contains the contents of a deployment found in a halconfig .deploymentConfigurations[0]
+    # spec.spinnakerConfig.config - This section contains the contents of a deployment found in a halconfig .deploymentConfigurations[0].
     config:
-      version: 1.17.1   # the version of Spinnaker to be deployed
+      version: 1.17.1   # - The version of Spinnaker that gets deployed.
       persistentStorage:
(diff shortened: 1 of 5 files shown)


doc/options.md
```diff
@@ -173,12 +173,11 @@ spec:
 ```
 
 ## `spec.expose`
-This section contains configuration for exposing Spinnaker. It is optional, if you omit it
-no load balancer will be created (or deleted if you remove it). 
+Optional. Controls how Spinnaker gets exposed. If you omit this section, no load balancer gets created. If this section is removed, the Load Balancer gets deleted.
```
german-muzquiz: If this section is removed, the Load Balancers are **not** deleted. This is because we want to support users to "opt out" of operator managed load balancers.

doc/uninstalling.md
```diff
@@ -1,8 +1,9 @@
 # Uninstalling the Operator
 
-If for some reason the operator needs to be uninstalled/deleted, there are still two ways in which Spinnaker itself can be prevented from being deleted, explained in the following sections.
+If Operator needs to be uninstalled or deleted, you can keep the Spinnaker deployments from being deleted in one of two ways: [replace Operator with Halyard](#replacing-the-operator-with-halyard) or [remove Operator ownership of Spinnaker resources](#removing-operator-ownership-of-spinnaker-resources).
```
german-muzquiz: Is the link `#replacing-the-operator-with-halyard` still correct? I see that the name of the section was slightly changed below to `Replacing Operator with Halyard`

doc/options.md
brian-armory: Changed the sentence to 

> If this section is removed, the Load Balancer does not deleted.

doc/uninstalling.md
brian-armory: Updated the verbs to match

PR #75, opened 2019-12-12
fix(validation/k8s): Use rawClient for secrets context

german-muzquiz:


pkg/controller/accountvalidating/accountvalidating_controller.go
```diff
@@ -83,3 +90,13 @@ func (v *accountValidatingController) InjectDecoder(d *admission.Decoder) error
 	v.decoder = d
 	return nil
 }
+
+// InjectConfig injects the rest config for creating raw kubernetes clients.
+func (v *accountValidatingController) InjectConfig(c *rest.Config) error {
```
ncknt: The `NewForConfig` isn't cheap. Instead of making a `clientSet` every reconciler loop, can we just keep the config and do a `client = corev1.NewForConfig(&config)` where we need it? We just need `client.Secret(namespace).Get(secretname)`.

PR #80, opened 2020-01-21
fix(patch): Changed patch strategy to three way merge

german-muzquiz:
* Using three way merge strategy for patching. The "third" payload is used specifically for deletions. Regular two-way merges are used for additions/changes.
* Using a raw client to avoid reading stale data from cache.
* Fixed an issue in integration tests where cluster role bindings collide when running tests in parallel

### DIFF
--- a/pkg/deploy/spindeploy/save.go
+++ b/pkg/deploy/spindeploy/save.go
@@ -3,7 +3,6 @@ package spindeploy
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/jsonmergepatch"
@@ -63,7 +62,7 @@ func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen
 func (d *Deployer) saveObject(ctx context.Context, obj runtime.Object, skipCheckExists bool, logger logr.Logger) error {
 	// Check if it exists
 	if !skipCheckExists {
-		if err := d.patch(obj); err != nil {
+		if err := d.patch(ctx, obj); err != nil {
 			logger.Error(err, fmt.Sprintf("Unable to save object: %v", obj))
 			return err
 		}
@@ -77,10 +76,10 @@ func (d *Deployer) deleteObject(ctx context.Context, obj runtime.Object) error {
 	return d.client.Delete(ctx, obj)
 }
 
-func (d *Deployer) patch(modifiedRaw runtime.Object) error {
+func (d *Deployer) patch(ctx context.Context, modifiedRaw runtime.Object) error {
 	modified, ok := modifiedRaw.(metav1.Object)
 	if !ok {
-		return errors.New(fmt.Sprintf("unable to save object %s because is not a metav1.Object", modifiedRaw.GetObjectKind().GroupVersionKind().String()))
+		return fmt.Errorf("unable to save object %s because is not a metav1.Object", modifiedRaw.GetObjectKind().GroupVersionKind().String())
 	}
 
 	gvk := modifiedRaw.GetObjectKind().GroupVersionKind()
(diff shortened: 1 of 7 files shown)


pkg/deploy/spindeploy/save.go
```diff
@@ -76,14 +77,14 @@ func (d *Deployer) deleteObject(ctx context.Context, obj runtime.Object) error {
 	return d.client.Delete(ctx, obj)
 }
 
-func (d *Deployer) patch(original runtime.Object) error {
-	o, ok := original.(metav1.Object)
+func (d *Deployer) patch(modifiedRaw runtime.Object) error {
+	modified, ok := modifiedRaw.(metav1.Object)
 	if !ok {
-		return errors.New("Unable to save object")
+		return errors.New(fmt.Sprintf("unable to save object %s because is not a metav1.Object", modifiedRaw.GetObjectKind().GroupVersionKind().String()))
```
ncknt: use `fmt.Errorf`

pkg/deploy/spindeploy/save.go
```diff
@@ -112,38 +113,53 @@ func (d *Deployer) patch(original runtime.Object) error {
...
+// createDeleteJson creates a json with potential fields to be removed from the original object
+func (d *Deployer) createDeleteJson(modifiedRaw runtime.Object, originalRaw runtime.Object) ([]byte, error) {
+	original, ok := originalRaw.(metav1.Object)
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("unable to cast %s to metav1.Object", originalRaw.GetObjectKind().GroupVersionKind().String()))
+	}
+	deleteObjectRaw := modifiedRaw.DeepCopyObject()
+	deleteObject, ok := deleteObjectRaw.(metav1.Object)
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("unable to cast %s to metav1.Object", deleteObjectRaw.GetObjectKind().GroupVersionKind().String()))
```
ncknt: `fmt.Errorf`

pkg/deploy/spindeploy/save.go
```diff
@@ -112,38 +113,53 @@ func (d *Deployer) patch(original runtime.Object) error {
...
 		Do().
-		Into(original)
+		Into(modifiedRaw)
+}
+
+// createDeleteJson creates a json with potential fields to be removed from the original object
+func (d *Deployer) createDeleteJson(modifiedRaw runtime.Object, originalRaw runtime.Object) ([]byte, error) {
+	original, ok := originalRaw.(metav1.Object)
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("unable to cast %s to metav1.Object", originalRaw.GetObjectKind().GroupVersionKind().String()))
```
ncknt: `fmt.Errorf`

pkg/deploy/spindeploy/save.go
```diff
@@ -112,38 +113,53 @@ func (d *Deployer) patch(original runtime.Object) error {
...
-	cp := original.DeepCopyObject()
+	originalRaw := modifiedRaw.DeepCopyObject()
 
-	err = i.Get().
-		Namespace(o.GetNamespace()).
-		Resource(rsc.Resource).
-		Name(o.GetName()).
-		Do().
-		Into(cp)
+	// avoid reading from cache
```
ncknt: Doesn't it read from the `client` cache on the contrary?

pkg/deploy/spindeploy/save.go
```diff
@@ -112,38 +113,53 @@ func (d *Deployer) patch(original runtime.Object) error {
...
+	err = d.client.Get(context.TODO(), types.NamespacedName{Namespace: modified.GetNamespace(), Name: modified.GetName()}, originalRaw)
```
ncknt: may be a good idea to explicitly pass the `context`

pkg/deploy/spindeploy/save.go
german-muzquiz: I'm not sure which client uses cache or not, I originally tested with the `RESTClient` produced by `Clientset` and it was reading a stale version of the resource that didn't have the latest changes (producing again an infinite loop). 
Then I changed to this other client in `Deployer` struct and it's reading the correct information every time.

PR #84, opened 2020-01-28
fix(files): Don't convert json files to yaml

german-muzquiz:
Don't try to sanitize secrets in json files like packer templates.


pkg/deploy/spindeploy/transformer/secrets.go
```diff
@@ -73,9 +73,15 @@ type kubernetesSecretCollector struct {
 // and file paths
 func (k *kubernetesSecretCollector) mapSecrets(svc string, secret *v1.Secret) error {
 	for key := range secret.Data {
+		// Is this a json string?
+		v := secret.Data[key]
+		s := strings.TrimSpace(string(v))
+		if s[0] == '{' && s[len(s)-1] == '}' {
```
ncknt: How about `[`?

german-muzquiz: 🤦‍♂ 

Thanks for finding that!

PR #86, opened 2020-02-07
doc(updates): Updated docs

german-muzquiz:


README.md
```diff
@@ -13,45 +24,30 @@ The validating admission controller [requires](https://kubernetes.io/docs/refere
...
-
 # Install or update CRDs cluster wide
 $ kubectl apply -f deploy/crds/
 
 # Install operator in namespace spinnaker-operator, see below if you want a different namespace
 $ kubectl create ns spinnaker-operator
 $ kubectl -n spinnaker-operator apply -f deploy/operator/cluster
 
-# Install Spinnaker in "spinnaker" namespace
+# Install Spinnaker in "spinnaker" namespace, connected to S3 bucket "spinnaker" (bucket must already exist and be accessible from Kubernetes EKS IAM role of the worker nodes)
```
ncknt: I'd remove comments about EKS as it sounds like a requirement to use the operator. Can we word it differently? Perhaps: `Note that this configuration uses an S3 bucket that Spinnaker needs to be able to access (e.g. on EKS, you can grant access to the worker node on which Spinnaker runs).

deploy/spinnaker/basic/spinnakerservice.yml
```diff
@@ -9,12 +9,10 @@ spec:
       persistentStorage:
         persistentStoreType: s3
         s3:
-          bucket: mybucket # Change me
+          bucket: spinnaker
```
ncknt: `spinnaker` sounds like a system thing that I shouldn't change. Can we either leave a note to change it or make it more obvious it's an example? Not everyone will know what this setting represents.

german-muzquiz: I'll leave a note that it can be changed safely. The idea here is to provide sensible defaults to avoid having users to change many settings to get started, just like we're creating the namespace `spinnaker-operator` and set it as default instead of having them to go an update the manifest to put a desired namespace.

kevinawoo: `spinnaker` won't work because s3 buckets are globally (like real real globally) unique

kevinawoo: how about `my-spinnaker-storage-bucket`? to make it more obvious? or `CHANGEMEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE`

deploy/spinnaker/complete/spinnakerservice.yml
```diff
@@ -11,8 +11,8 @@ spec:
       persistentStorage:
         persistentStoreType: s3
         s3:
-          bucket: mybucket
-          rootFolder: front50 # Change me
+          bucket: change-me # Change to a unique name. Spinnaker stores application and pipeline definitions here
```
aimeeu: The Docs team is moving to angle brackets for [placeholders](https://kubernetes.io/docs/contribute/style/style-guide/#use-angle-brackets-for-placeholders).  So how about `<change-me>`... or `<globally-unique-bucket-name>`

deploy/spinnaker/kustomize/spinnakerservice.yml
```diff
@@ -7,11 +7,11 @@ spec:
   spinnakerConfig:
     # spec.spinnakerConfig.config - This section contains the contents of a deployment found in a halconfig .deploymentConfigurations[0]
     config:
-      version: 1.15.1   # the version of Spinnaker to be deployed
+      version: 1.17.1   # the version of Spinnaker to be deployed
       persistentStorage:
         persistentStoreType: s3
         s3:
-          bucket: mybucket
+          bucket: change-me # Change to a unique name. Spinnaker stores application and pipeline definitions here
```
aimeeu: The Docs team is moving to angle brackets for [placeholders](https://kubernetes.io/docs/contribute/style/style-guide/#use-angle-brackets-for-placeholders).  So how about `<change-me>`... or `<globally-unique-bucket-name>`

aimeeu:
Thanks @german-muzquiz !  LGTM

PR #88, opened 2020-02-12
chore(validations): Better error message on validation timeout

german-muzquiz:
Provide a warning message before webhook admission request times out, telling the user how to disable individual validations. Example output in operator logs:
```
{"level":"info","ts":1581614409.39606,"logger":"spinvalidate","msg":"WARNING: Validations are taking too much time. Consider disabling individual validations in: 'spec.validations.providers.[name].enabled'. Providers: [appengine aws azure cloudfoundry dcos docker ecs google oracle], 'spec.validations.persistentStorage.[name].enabled'. Persistence storages: [azs gcs oracle s3], 'spec.validations.canary.[name].enabled'. Canaries: [aws google prometheus], 'spec.validations.pubsub.[name].enabled'. PubSubs: [google]"...
```


pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -90,6 +95,26 @@ func (v *spinnakerValidatingController) Handle(ctx context.Context, req admissio
 	return admission.ValidationResponse(true, "")
 }
 
+func (v *spinnakerValidatingController) warnInValidationDelay(d time.Duration, svc v1alpha2.SpinnakerServiceInterface) chan bool {
+	ch := make(chan bool, 1)
```
ncknt: consider a `chan struct{}` instead of bool

pkg/controller/spinnakervalidating/spinnakervalidating_controller.go
```diff
@@ -90,6 +95,26 @@ func (v *spinnakerValidatingController) Handle(ctx context.Context, req admissio
...
+	go func() {
+		select {
+		case <-ch:
+			return
+		case <-time.After(d):
+			validations := halyard.GetValidationKeys()
+			msg := fmt.Sprintf("WARNING: Validations are taking too much time. Consider disabling individual validations in: "+
```
ncknt: To be even more useful, can we output the name of the validations that are still running? This would probably need to be moved under the `validate` probably under `parallel.go` but that would be for the best:
- controller code would remain lean and clean
- we know what validations are still running
- we already process things in parallel so it'd integrate nicely should we want to have the options to cancel pending validations in the future.

german-muzquiz: Given that the process for creating the dynamic list of validators to run can also face some delays (like decrypting any remote secret), I think it's better to wrap the timeout around the full validations call in the controller, but we can still have the logic of which validations are currently running in `validate`.

german-muzquiz: After some inspection of parallel validator, outputting the running validator names will need a bit more code changes than I would like for a patch release.

PR #89, opened 2020-02-13
feat(release): Github action for patch releases

german-muzquiz:
Github action to help with automation of patch releases.

This is meant to be triggered after CI passes and docker image for the new release is pushed to registry. CI would need to make a HTTP request to github dispatch endpoint for this repository, indicating the new image tag that must follow semver version format (x.y.z).

This action will then do the following:
- Make sure a release branch already exists for this version (i.e. version=0.1.0, branch=release-0.1.x)
- Make sure the version has not already been released (by checking if the tag already exists)
- Update docker image tag in operator `deployment.yml` files
- Compress manifest files
- Create the github release (which creates automatically the tag)
- Add manifests to release

### DIFF
--- a/.github/workflows/patch-release.yml
+++ b/.github/workflows/patch-release.yml
@@ -1,0 +1,49 @@
+on:
+  repository_dispatch:
+    types: [patch-release]
+
+jobs:
+  patch-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.client_payload.branch }}
+      - name: Validate branch and fetch tags
+        run: |
+          [[ ${{ github.event.client_payload.branch }} != release-* ]] && echo "Branch ${{ github.event.client_payload.branch }} is not a release branch" && exit 1
+          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+      - name: Validate tag
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,51 @@
...
+          ref: ${{ steps.branch_step.outputs.branch }}
+      - name: Fetch tags
+        run: |
+          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+      - name: Validate tag
+        run: |
+          git tag v${{ github.event.client_payload.version }}
+      - name: Update manifests version
+        run: |
+          cat deploy/operator/basic/deployment.yaml | sed "s|image: armory/spinnaker-operator:.*|image: armory/spinnaker-operator:${{ github.event.client_payload.version }}|" | sed "s|imagePullPolicy:.*|imagePullPolicy: IfNotPresent|" > deploy/operator/basic/deployment.yaml.new
```
kevinawoo: not `always`?

kevinawoo: oh well.... i guess since it's tagged, it should never change. OK!

german-muzquiz: `Always` is used for dev work because we can update the same tag (`dev`) over and over. For releases we use unique tags for each release, so `imagePullPolicy` can be changed to `IfNotPresent`.

PR #90, opened 2020-02-14
fix(operator): Update from extensions/v1beta1/deployment to apps/v1/deployment (WIP)

justinrlee:


justinrlee:
I'm trying to figure out how to build and test this.

justinrlee:
btw - this works, but the halyard that comes with operator is still generating manifests with v1beta2, which was fixed here, I think: https://github.com/spinnaker/halyard/pull/1443

PR #95, opened 2020-02-26
feat(operator/telemetry): Add ability to report deploy method via telemetry

castanedaci:
This change allows sending the telemetry information with deployment method and version of spinnaker operator.

There was a new Transformer added that will perform the update in properties inside of `echo.yml ` secret 

```
telemetry.deploymentMethod.type
telemetry.deploymentMethod.version
```

Also, we are generating a MANIFEST file to persist operator build information + version

```
Spinnaker-Operator-Version=0.3.0-1d90342-dirty
Built-By=Jenkins
Build-Date=2020-02-27_18:55:06
Branch=master
Revision=1d90342
Build-Go-Version=go version go1.13.6 darwin/amd64

```


pkg/deploy/spindeploy/transformer/telemety.go
```diff
@@ -0,0 +1,114 @@
+package transformer
```
german-muzquiz: Typo in file name: `telemetry.go` instead of `telemety.go`

version/version.go
```diff
@@ -1,5 +1,17 @@
...
-	Version = "0.0.1"
+	Version = getVersion()
 )
+
+func getVersion() string {
+	version, err := ioutil.ReadFile("./operator-version")
+	if err != nil {
+		return "Unknown"
+	}
+	return string(version)
```
german-muzquiz: Let's take into account the case where the version file is something like this:
```
0.3.0 

```
You need to take only the first line, and do a `trim` to delete whitespaces around the version number.

version/version.go
```diff
@@ -1,5 +1,17 @@
 package version
 
+import (
+	"io/ioutil"
+)
+
 var (
-	Version = "0.0.1"
+	Version = getVersion()
```
german-muzquiz: It's better to invoke `getVersion()` from the init function of the package:
```
var Version string
func init() {
    b, err := ioutil.ReadFile("./operator-version")
    ...
}

File: pkg/deploy/spindeploy/transformer/telemety.go [hunk 4]
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/armory/spinnaker-operator/version"
+	"github.com/ghodss/yaml"
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	EchoConfigFile     = "echo.yml"
```
german-muzquiz: If these constants are not used outside of this file, you can make them private by starting with lowercase (`echoConfigFile` instead of `EchoConfigFile`).

pkg/deploy/spindeploy/transformer/telemety.go
```diff
@@ -0,0 +1,114 @@
...
+	return "Telemetry"
+}
+
+func (t *telemetryTransformer) TransformConfig(ctx context.Context) error {
+	return nil
+}
+
+func (t *telemetryTransformer) TransformManifests(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig) error {
+	for _, cfg := range gen.Config {
+		for k := range cfg.Resources {
```
german-muzquiz: I think it would be better to explicitly search for `echo` service config here, and once found iterate only echo's Resources instead of iterating the resources of all services:
```
for svc, cfg := range gen.Config {
    if svc != "echo" {
        continue
    }
    for k := range cfg.Resources {
    ...
```

pkg/deploy/spindeploy/transformer/telemety.go
```diff
@@ -0,0 +1,114 @@
...
+
+// mapTelemetrySecret goes through EchoConfigFile and set deployment method information
+func mapTelemetrySecret(secret *v1.Secret) error {
+	for key := range secret.Data {
+		if EchoConfigFile == key {
+
+			// Attempt to deserialize as YAML
+			m := make(map[string]interface{})
+			if err := yaml.Unmarshal(secret.Data[key], &m); err != nil {
+				continue
```
german-muzquiz: If we know that `secret.Data[key]` is `echo.yml` and still get an error unmarshalling, I think we need to return that error here instead of continuing the loop.

pkg/deploy/spindeploy/transformer/telemety.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			data, err := setTelemetryDeploymentMethod(m)
+			if err != nil {
+				continue
```
german-muzquiz: Same here about throwing the error instead of continuing the loop.

pkg/deploy/spindeploy/transformer/telemety.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type Telemetry struct {
+	Enabled                 bool             `json:"enabled"`
+	Endpoint                string           `json:"endpoint"`
+	InstanceId              string           `json:"instanceId"`
+	SpinnakerVersion        string           `json:"spinnakerVersion"`
+	DeploymentMethod        DeploymentMethod `json:"deploymentMethod"`
+	ConnectionTimeoutMillis int              `json:"connectionTimeoutMillis"`
+	ReadTimeoutMillis       int              `json:"readTimeoutMillis"`
```
german-muzquiz: I think we can update `DeploymentMethod` only without having to declare all available fields inside `Telemetry`. The problem is that we would need to change the operator project every time a new field is added to telemetry, which can be avoided.

pkg/deploy/spindeploy/transformer/telemety.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (t *telemetryTransformerGenerator) NewTransformer(svc v1alpha2.SpinnakerServiceInterface,
+	client client.Client, log logr.Logger) (Transformer, error) {
```
dogonthehorizon: Is `NewTransformer` part of an interface? Can we document why we will never return `error` here?

pkg/deploy/spindeploy/transformer/telemety.go
castanedaci: I thought the same at the beginning, I just add them thinking about having all the fields available for future changes.

pkg/deploy/spindeploy/transformer/telemety.go
castanedaci: Yes, it is, the interface is coming from `transformer.Genereator`.
Since we are not performing anything than just initializing telemetryTransformer, we don't have to worry about any error so far.

pkg/version/manifest.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+const (
+	operatorHomePath = "OPERATOR_HOME"
+	manifestFile     = "MANIFEST"
+)
+
+var manifest = make(map[string]string)
+
+func init() {
```
ncknt: There's little guarantee for the `init()` method - especially with two such methods in the same module. Instead of reading in init(), can you just lazily read it? First time, `GetOperationVersion` is called, read the manifest and keep it locally.

pkg/version/version.go
```diff
@@ -0,0 +1,11 @@
+package version
+
+var SpinnakerOperatorVersion string
+
+func init() {
+	v := manifest["Spinnaker-Operator-Version"]
```
ncknt: I think we can just call it `Version` :)

pkg/version/manifest.go
ncknt: That would also let you return error in case the build was messed up vs panic.

build/generate-manifest.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ -z "$RELEASE_VERSION" ]; then
+  SPINNAKER_OPERATOR_VERSION="$SPINNAKER_OPERATOR_VERSION-$VERSION"
+fi
+
+echo "Spinnaker-Operator-Version="$SPINNAKER_OPERATOR_VERSION > build/MANIFEST
+echo "Built-By="$(whoami) >> build/MANIFEST
+echo "Build-Date="$(date +'%Y-%m-%d_%H:%M:%S') >> build/MANIFEST
+echo "Branch="$(git rev-parse --abbrev-ref HEAD) >> build/MANIFEST
+echo "Revision="$(git describe --always) >> build/MANIFEST
+echo "Build-Go-Version="$(go version) >> build/MANIFEST
```
ncknt: I like that! Good idea.

pkg/deploy/spindeploy/transformer/telemetry.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+func (t *telemetryTransformer) TransformManifests(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig) error {
+	for svc, cfg := range gen.Config {
+		if svc != "echo" {
+			continue
+		}
+		for k := range cfg.Resources {
+			sec, ok := cfg.Resources[k].(*v1.Secret)
```
ncknt: We unfortunately cannot offer any guarantee that this is the secret you want and you may have more than one. You should use `secretName := util.GetMountedSecretNameInDeployment(config.Deployment, k, "/opt/spinnaker/config")` followed by `sec := getSecretFromConfig(cfg.Resources[k], secretName)`.

Check out `accounts.go` for an example.

pkg/deploy/spindeploy/transformer/telemetry.go
```diff
@@ -0,0 +1,111 @@
...
+			if ok {
+				err := mapTelemetrySecret(sec)
+				if err != nil {
+					return err
```
ncknt: Telemetry shouldn't halt deployment. If it fails, just log it.

pkg/version/manifest.go
```diff
@@ -0,0 +1,36 @@
...
+	path := os.Getenv(operatorHomePath)
+	body, err := ioutil.ReadFile(path + "/" + manifestFile)
+
+	if err != nil {
+		return
+	}
+
+	raws := strings.Split(string(body), "\n")
+
+	for _, raw := range raws {
```
ncknt: Can you add some tests?

pkg/deploy/spindeploy/transformer/telemetry.go
```diff
@@ -0,0 +1,111 @@
...
+				return err
+			}
+
+			secret.Data[key] = data
+		}
+	}
+	return nil
+}
+
+func setTelemetryDeploymentMethod(row map[string]interface{}) ([]byte, error) {
```
ncknt: This needs some test. What if there's a telemetry key? What if there's not? What if there's already a `telemetry.type`? etc.

PR #97, opened 2020-03-06
chore(interfaces): Use interfaces instead of versioned packages

german-muzquiz:
In order to support easy CRD version changes, now the code base uses version agnostic interfaces instead of versioned packages. Versioned packages implement these interfaces, and the goal is that every time a new CRD version is added, we start with copying the previous version package files.

```
Interfaces <----- v1alpha2
      ^-------- v1beta1
```

An example of how the code changes, from:
```
spinSvc.Spec.SpinnakerConfig
```
To:
```
spinSvc.GetSpec().getSpinnakerConfig()
```
All integration tests are passing after these changes.


german-muzquiz:
Probably the most important files in this PR are `interfaces.go`and `spinnakerservice_functions.go`

pkg/apis/spinnaker/interfaces/interfaces.go
```diff
@@ -0,0 +1,377 @@
...
+	NewKubernetesAuth() KubernetesAuth
+	NewHashStatus() HashStatus
+	NewExposeConfigServiceOverrides() ExposeConfigServiceOverrides
+	NewSpinnakerValidation() SpinnakerValidation
+	NewValidationSetting() ValidationSetting
+	GetGroupVersion() schema.GroupVersion
+	DeepCopyLatestTypesFactory() LatestTypesFactory
+}
+
+type TypesFactory interface {
```
ncknt: Shouldn't we be using conversions instead? As in https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/ where we'd register a conversion function into the scheme and only worry about the latest?

PR #101, opened 2020-03-13
feat(stats): Rename telemetry to stats due to change in halyard

castanedaci:
Rename telemetry to stats due to spinnaker/halyard#1565

### DIFF
--- a/pkg/deploy/spindeploy/transformer/stats_test.go
+++ b/pkg/deploy/spindeploy/transformer/stats_test.go
@@ -9,7 +9,7 @@ import (
 	"testing"
 )
 
-func TestMapTelemetrySecret(t *testing.T) {
+func TestMapStatsSecret(t *testing.T) {
 
 	// given
 	secretContent := `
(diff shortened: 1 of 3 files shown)


pkg/deploy/spindeploy/transformer/stats_test.go
```diff
@@ -19,20 +19,20 @@ metadata:
   name: spin-echo-files-287979322
   type: Opaque
 data:
-  echo.yml: dGVsZW1ldHJ5OgogIGVuYWJsZWQ6IHRydWUKICBkZXBsb3ltZW50TWV0aG9kOgogICAgdHlwZTogaGFseWFyZAogICAgdmVyc2lvbjogMS4zMi4wLTQz
+  echo.yml: c3RhdHM6CiAgZW5hYmxlZDogdHJ1ZQogIGRlcGxveW1lbnRNZXRob2Q6CiAgICB0eXBlOiBoYWx5YXJkCiAgICB2ZXJzaW9uOiAxLjMyLjAtNDM=
 `
 	data := make(map[string][]byte)
 
 	secret := &v1.Secret{Data: data}
 	_ = yaml.Unmarshal([]byte(secretContent), secret)
```
german-muzquiz: I know this wasn't changed by this PR, but could you check for the returned error instead of ignoring? Something like
```
if !assert.Nil(t, yaml.Unmarshal([]byte(secretContent), secret)) {
    return
}
```

PR #107, opened 2020-04-02
feat(jsonpatches): support for merge, 6902, and strategic merge patches

ncknt:


pkg/deploy/spindeploy/transformer/patches.go
```diff
@@ -0,0 +1,204 @@
...
+		if err != nil {
+			return nil, err
+		}
+		bytes, err := patchObj.Apply(originalJS)
+		// TODO: This is pretty hacky, we need a better structured error from the json-patch
+		if err != nil && strings.Contains(err.Error(), "doc is missing key") {
+			msg := err.Error()
+			ix := strings.Index(msg, "key:")
+			key := msg[ix+5:]
+			return bytes, fmt.Errorf("Object to be patched is missing field (%s)", key)
```
german-muzquiz, line 185: It's probably good to wrap the original error to provide more context:
```
return bytes, fmt.Errorf("Object to be patched is missing field (%s). Original error: %w", key, err)
```
https://blog.golang.org/go1.13-errors (Wrapping errors with %w)

pkg/deploy/spindeploy/transformer/patches.go
```diff
@@ -0,0 +1,204 @@
...
+		return bytes, err
+
+	case types.MergePatchType:
+		return jsonpatch.MergePatch(originalJS, patchJS)
+
+	case types.StrategicMergePatchType:
+		// get a typed object for this GVK if we need to apply a strategic merge patch
+		obj, err := creater.New(gvk)
+		if err != nil {
+			return nil, fmt.Errorf("cannot apply strategic merge patch for %s locally, try --type merge", gvk.String())
```
german-muzquiz, line 196: I don't understand the error message indicating `try --type merge`, is this message from a CLI tool? How does it look like in the `SpinnakerService` manifest?

pkg/deploy/spindeploy/transformer/patches.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		return strategicpatch.StrategicMergePatch(originalJS, patchJS, obj)
+
+	default:
+		// only here as a safety net - go-restful filters content-type
+		return nil, fmt.Errorf("unknown Content-Type header for patch: %v", patchType)
```
german-muzquiz, line 202: I think a better error message would be something related with `SpinnakerService` manifest, like `unknown patch strategy x`

pkg/deploy/spindeploy/transformer/patches.go
```diff
@@ -0,0 +1,204 @@
...
+	return nil, errors.New("not parsed as a deployment, patching failed")
+}
+
+func (p *patchTransformer) asService(data []byte) (*v1.Service, error) {
+	dser := scheme.Codecs.UniversalDecoder()
+	svc := &v1.Service{}
+	obj, _, err := dser.Decode(data, &schema.GroupVersionKind{
+		Group:   "apps",
+		Version: "v1",
+		Kind:    "Deployment",
```
german-muzquiz, line 144: Isn't this gvk for service `v1.Service` as is line 162?

pkg/deploy/spindeploy/transformer/patches_test.go
```diff
@@ -0,0 +1,164 @@
...
+				}
+			},
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			_, spinsvc := th.setupTransformerFromSpinText(&exposeLbTransformerGenerator{}, c.spinsvc, t)
+			gen := &generated.SpinnakerGeneratedConfig{}
+			test.AddDeploymentToGenConfig(gen, "gate", "testdata/input_deployment.yml", t)
+			p := patchTransformer{svc: spinsvc}
```
german-muzquiz, line 157: I think there's no need for this line, since the above method call `th.setupTransformerFromSpinText` in line 154 already produces the instance of the transformer to be used, just provide the `patchTransformerGenerator`, which by the way, implicitly tests the generator itself that seemed to have a bug ;)

pkg/deploy/spindeploy/transformer/patches.go
```diff
@@ -0,0 +1,204 @@
...
+	svc interfaces.SpinnakerService
+	log logr.Logger
+}
+
+type patchTransformerGenerator struct{}
+
+func (g *patchTransformerGenerator) NewTransformer(svc interfaces.SpinnakerService,
+	client client.Client, log logr.Logger) (Transformer, error) {
+
+	tr := serverPortTransformer{svc: svc, log: log}
```
german-muzquiz, line 35: Shouldn't this instantiate `patchTransformer` instead?

pkg/deploy/spindeploy/transformer/patches.go
ncknt: Yes but it still works. Makes me wonder how...

pkg/deploy/spindeploy/transformer/patches.go
ncknt: This is how kustomize does it: error is swallowed because the underlying one doesn't make sense to the user. I'd rather keep it that way.

ncknt:
We could also remove `spec.expose.annotations` but that would mean a bump in version and perhaps a conversion (or at least some documentation). I'll keep it separate for now but we probably need to address before making a release.

PR #112, opened 2020-04-22
feat(accounts/dynamic): use dynamic-config if Spinnaker is 1.19+ and …

ncknt:
…mount to /opt/spinnaker/config/dynamic


ncknt:
Closing for now. We're waiting for a better method to reflect change in Clouddriver and in all services in general.

PR #113, opened 2020-04-22
fix(status): Watch spinnaker service changes to update status

german-muzquiz:
Sometimes, if the service load balancers don't get assigned a public IP before all spinnaker pods are up and running, after load balancers are ready there's no reconciliation loop, causing that commands like
```
kubectl describe spinsvc <spinnaker>
```
To don't report UI and API urls.

Updated Halyard version fixes a race condition where parsing the BOM file sometimes failed when using one operator to deploy multiple spinnakers at the same time.

### DIFF
--- a/halyard-version
+++ b/halyard-version
@@ -1,1 +1,1 @@
-operator-0.4.x
+operator-75233080
--- a/pkg/controller/spinnakerservice/spinnakerservice_controller.go
+++ b/pkg/controller/spinnakerservice/spinnakerservice_controller.go
@@ -10,6 +10,7 @@ import (
 	"github.com/armory/spinnaker-operator/pkg/secrets"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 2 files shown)


halyard-version
```diff
@@ -1 +1 @@
-operator-0.4.x
+operator-75233080
```
ncknt, line 1: Will Halyard be retagged by the build process?

pkg/controller/spinnakerservice/spinnakerservice_controller.go
```diff
@@ -68,7 +69,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	// Watch for potential object owned by SpinnakerService
-	return c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    TypesFactory.NewService(),
+	})
+	if err != nil {
+		return err
+	}
+	return c.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{
```
ncknt, line 79: If I get that right, it should mean users lose the ability to edit service after the fact? (e.g. to change the type). It's probably okay since we now support patches but worth calling out in the change log

halyard-version
german-muzquiz: There's a Jenkins job for Halyard that generates this type of tag (`operator-<commit hash>`) and also retags to `operator-dev`. 
For development, this repository uses the manifest files in `deploy/operator/...` which point to `operator-dev` tag. During operator build process, Halyard image tag is replaced by the value in this `halyard-version` file.
I have seen that the downside of having hayard version in format `0.4.x` is that is difficult to figure out the next version of Halyard for an operator patch update, so I thought that having the commit hash would be enough.

pkg/controller/spinnakerservice/spinnakerservice_controller.go
german-muzquiz: That's right when we're exposing spinnaker. Although I haven't tested the scenario, if you omit the `expose` section entirely you can still make manual changes to the services, operator will enter the reconciliation loop, look that expose is disabled and will not touch the services. However, manifests generated by halyard may override external changes anyway, so using patches is the way to get them customized.

PR #116, opened 2020-04-27
feat(defaults): Add transform for profile default values

nhtzr:
Add `archaius.enabled: false` as a default in the java services if they are not defined in the profile.

### DIFF
--- a/pkg/deploy/spindeploy/transformer/defaults.go
+++ b/pkg/deploy/spindeploy/transformer/defaults.go
@@ -1,0 +1,99 @@
+package transformer
+
+import (
+	"context"
+	"fmt"
+	"github.com/armory/spinnaker-operator/pkg/apis/spinnaker/interfaces"
+	"github.com/armory/spinnaker-operator/pkg/generated"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// defaultsTransformer inserts default values to *-local profile to each service
+type defaultsTransformer struct {
+	svc    interfaces.SpinnakerService
+	log    logr.Logger
+	client client.Client
+}
+
+type defaultsTransformerGenerator struct{}
+
+func (g *defaultsTransformerGenerator) GetName() string {
+	return "Defaults"
+}
+
+func (a *defaultsTransformerGenerator) NewTransformer(
+	svc interfaces.SpinnakerService,
+	client client.Client,
+	log logr.Logger) (Transformer, error) {
+	return &defaultsTransformer{svc: svc, log: log, client: client}, nil
+}
+
+func (a *defaultsTransformer) TransformConfig(ctx context.Context) error {
+	config := a.svc.GetSpinnakerConfig()
+	for profileName, p := range config.Profiles {
+		handled, err := SetArchaiusDefaults(p, profileName)
+		if handled {
+			a.log.Info("Archaius defaults: Applied to %s, errors %e", profileName, err)
+		} else {
+			a.log.Info("Archaius defaults: Skipped %s", profileName)
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


pkg/deploy/spindeploy/transformer/defaults.go
```diff
@@ -0,0 +1,99 @@
...
+	archaius["enabled"] = false
+	a.log.Info("Archaius defaults: Applied to %", profileName)
+	return nil
+}
+
+func (a *defaultsTransformer) TransformManifests(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig) error {
+	return nil // noop
+}
+
+func isJavaService(profileName string) bool {
```
german-muzquiz, line 77: This function can be moved to `constants`, and converted to a map or array for shorter lookup:
```
var javaServices = map[string]bool{"echo": true, "front50": true}

func IsJavaService(svcName string) {
	_, ok := javaServices[svcName]
	fmt.Printf("echo is java service: %t", ok)
}
```

pkg/util/constants.go
```diff
@@ -11,3 +11,34 @@ const (
...
+		"fiat":        {},
+		"igor":        {},
+		"rosco":       {},
+		"front50":     {},
+		"kayenta":     {},
+		"gate":        {},
+		"dinghy":      {},
+		"terraformer": {},
+	}
+	SpinnakerJavaServices = map[string]struct{}{
```
german-muzquiz, line 29: I would suggest having a single map instead of a map per service type. Something like this:
```
var (
	SpinnakerServices = map[string]SpinSvcMetadata{
		"clouddriver": {
			Name:       "clouddriver",
			Type:        "java",
		},
         }
)

type SpinSvcMetadata struct {
	Name       string
	Type       string
}
```
Along with appropriate methods in this file to fecth all service names, all java services and so on.

pkg/util/constants.go
```diff
@@ -11,3 +11,34 @@ const (
...
+		"orca":        {},
+		"echo":        {},
+		"fiat":        {},
+		"igor":        {},
+		"rosco":       {},
+		"front50":     {},
+		"kayenta":     {},
+		"gate":        {},
+	}
+	SpinnakerGoServices = map[string]struct{}{
```
german-muzquiz, line 40: OSS spinnaker doesn't have any Go services. Dinghy and terraformer are part of Armory Spinnaker distribution, so they should not be included here.

pkg/deploy/spindeploy/transformer/defaults.go
```diff
@@ -0,0 +1,96 @@
...
+		err := a.setArchaiusDefaults(p, profileName)
+		if err != nil {
+			return fmt.Errorf("found error while handling profile %s: %e", profileName, err)
+		}
+	}
+	return nil
+}
+
+func (a *defaultsTransformer) setArchaiusDefaults(profile interfaces.FreeForm, profileName string) error {
+	if !isJavaService(profileName) {
```
german-muzquiz, line 48: See comment about having a method in utils returning all java service names, we can iterate those for adding profiles since the beginning and remove this check.

pkg/deploy/spindeploy/transformer/defaults_test.go
```diff
@@ -0,0 +1,90 @@
...
+`
+	tr, spinsvc := th.setupTransformerFromSpinText(&defaultsTransformerGenerator{}, s, t)
+	before_ := spinsvc.GetSpinnakerConfig().Profiles["gate"]
+	before := *before_.DeepCopy()
+	err := tr.TransformConfig(context.TODO())
+	assert.Nil(t, err)
+
+	config := spinsvc.GetSpinnakerConfig()
+	gate := config.Profiles["gate"]
+	assert.NotNil(t, gate)
```
german-muzquiz, line 28: I think this needs to check that all services have a profile section with `archaius: true`, not just gate.

PR #121, opened 2020-05-12
fix(kustomize): detect change in kustomize patches

ncknt:


german-muzquiz:
This is almost identical to `config.go` change detector. Could it be better to modify that one to hash the entire `spec` section? As it is today, I think changes in `validation`, `profiles` or `service-settings` sections will not trigger the transformers either.

ncknt:
Sure I can take a look. I don't expect a change in `validation` to trigger anything.

PR #131, opened 2020-05-22
fix(secrets): k8s secrets decrypted in halyard

german-muzquiz:
* Mount k8s secrets env vars and volumes in `monitoring-daemon` containers
* Updated integration tests

### DIFF
--- a/integration_tests/integration_test.go
+++ b/integration_tests/integration_test.go
@@ -76,8 +76,6 @@ func TestKubernetesAndUpgradeOverlay(t *testing.T) {
 		Account{Name: "dockerhub", Type: "dockerRegistry"},
 		Account{Name: "kube-sa", Type: "kubernetes"},
 		Account{Name: "kube-file-reference", Type: "kubernetes"})
-	e.VerifyAccountsExist("/artifacts/credentials", t,
-		Account{Name: "dockerhub", Types: []string{"s3/object"}})
 	if t.Failed() {
 		return
 	}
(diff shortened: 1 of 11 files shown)


integration_tests/integration_test.go
```diff
@@ -72,6 +73,7 @@ func TestKubernetesAndUpgradeOverlay(t *testing.T) {
 
 	// verify accounts
 	e.VerifyAccountsExist("/credentials", t,
+		Account{Name: "dockerhub", Type: "dockerRegistry"},
```
cristhian-castaneda, line 76: Thanks for that, I forgot to add this.

integration_tests/testdata/spinnaker/overlay_secrets/spinnaker-template.yml
```diff
@@ -38,6 +38,13 @@ spec:
               omitKinds: []
               onlySpinnakerManaged: false
           primaryAccount: kube-s3-secret
+      metricStores:
+        datadog:
+          enabled: true
+          api_key: encrypted:k8s!n:testsecrets!k:datadog-apikey
+          app_key: encrypted:s3!b:{{.S3Bucket}}!f:secrets/secrets.yml!r:{{.S3BucketRegion}}!k:datadog.appkey
```
cristhian-castaneda, line 45: Should we keep both?

cristhian-castaneda: Sorry I didn't read correctly.

PR #135, opened 2020-06-02
feat(util): Util functions for manifest manipulation

german-muzquiz:


pkg/util/manifests.go
```diff
@@ -0,0 +1,83 @@
...
+func addSecretMountIfNeeded(volumeName, mountPath string, d *appsv1.Deployment, filter func(c v1.Container) bool) {
+	ctrs := make([]v1.Container, 0)
+	for _, c := range d.Spec.Template.Spec.Containers {
+		if !filter(c) {
+			ctrs = append(ctrs, c)
+			continue
+		}
+		added := false
+		for _, vm := range c.VolumeMounts {
+			if vm.Name == volumeName && vm.MountPath == mountPath {
```
ncknt, line 73: Shouldn't it be an error to have two volumes mounted to the same path with different volume name? Either return the volume name to be use or allow the method to error.

pkg/util/manifests.go
```diff
@@ -0,0 +1,83 @@
...
+			ctrs = append(ctrs, c)
+			continue
+		}
+		found := false
+		vars := make([]v1.EnvVar, 0)
+		for _, ce := range c.Env {
+			if ce.Name == e.Name {
+				found = true
+				if appendIfExists {
+					ce.Value = fmt.Sprintf("%s%s", ce.Value, e.Value)
```
ncknt, line 23: If the env var exists, I can see how I'd want to control how it's appended.
e.g. Adding a `test` Spring profile should set the value to `test` if it doesn't exist but should append `,test` otherwise.

pkg/util/manifests.go
german-muzquiz: You're right, I'll handle that as an error, maybe there's a configmap already mounted in the same path where we want to mount a secret

ncknt: What if you just want to update the secret? I believe we already have similar code.

german-muzquiz: If the secret has the same name and mount path this method would be noop, but if it has a different name then we need to delete it to add the new one instead. What about a bool param to indicate if we should replace any mounted volume?
```
func AddVolumeMountAsSecret(d *appsv1.Deployment, s *v1.Secret, mountPath string, replace bool, filter func(c v1.Container) bool) error {
}
```

ncknt: I wouldn't delete an existing secret implicitly like that. If we pass an existing volume name with a different mount path -> fine, we can mount twice (I believe?). If we mount two volumes to the mount path, it might actually work if the keys are specified and don't overlap (not sure either!) but it's probably not what was intended.

Maybe this is not a good candidate for a util function after all, and all we need is a `getVolumeMount(containername string, d *appsv1.Deployment) []v1.VolumeMount`, it's not worth it IMO.

german-muzquiz: I see adding volumes as a common operation (we do that for k8s secrets and maybe also dynamic accounts). Maybe it's difficult to accomodate for a common enough method, I'll just leave the env var util method.

PR #146, opened 2020-07-06
fix(validation): Issue patch to update status

ncknt:


german-muzquiz:
What's this adressing? Is it a bug only hapenning with account CRDs?

ncknt:
It should fix the error given by the validation controller when trying to update the validation hashes. That error happens when the controller's cache is out of sync. It sends a `PUT` based on the UID that is now out of date and the server correctly rejects it. It's okay during a reconciliation loop - not during validation. This method uses patch. It should also let us remove hashes for accounts no longer used (not included in this PR)

pkg/validate/accounts.go
```diff
@@ -93,7 +107,18 @@ func (a *accountValidator) Validate(spinSvc interfaces.SpinnakerService, options
 	if err != nil {
 		return NewResultFromError(fmt.Errorf("Validator for account '%s' detected an error:\n  %w", a.name, err), a.fatal)
 	}
-	return ValidationResult{}
+	p := getHashPatch(a.key, a.hash, a.t)
+	return ValidationResult{
+		StatusPatches: []jsonpatch.JsonPatchOperation{*p},
```
german-muzquiz, line 112: Should this be added to the rest of validators as well (not just account validators)?
For example, if I read correctly only `account-Kubernetes-spinnaker` is included in this logic:
```
status:
  apiUrl: http://localhost:8084
  lastDeployed:
    account-Kubernetes-spinnaker:
      hash: 37a6259cc0c1dae299a7866489dff0bd
      lastUpdatedAt: "2020-07-01T17:46:47Z"
    config:
      hash: ddf37a5ab2eae8494f76da12869ba970
      lastUpdatedAt: "2020-07-06T19:46:38Z"
    kustomize:
      hash: 37a6259cc0c1dae299a7866489dff0bd
      lastUpdatedAt: "2020-07-06T19:46:38Z"
  serviceCount: 8
  services:
  ...
```

ncknt: As far I can tell, that's the only place where we update the status on the validation controller.

PR #168, opened 2020-08-27
feat(health-check): check spinnaker status

cristhian-castaneda:


pkg/controller/spinnakerservice/spinnakerservice_controller.go
```diff
@@ -139,10 +140,10 @@ func (r *ReconcileSpinnakerService) Reconcile(request reconcile.Request) (reconc
 			return reconcile.Result{Requeue: true}, nil
 		}
 	}
-	sc := newStatusChecker(r.client, reqLogger, TypesFactory)
+	sc := newStatusChecker(r.client, reqLogger, TypesFactory, r.evtRecorder, util.NewK8sLookup(r.client))
 	if err = sc.checks(instance); err != nil {
 		r.evtRecorder.Eventf(instance, corev1.EventTypeWarning, "StatusError", "Error updating SpinnakerService status: %s", err.Error())
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true}, err
```
ncknt, line 146: You shouldn't need to requeue. If an error is returned, the operator framework will do a retry with back-off to avoid infinite loops.

integration_tests/integration_test.go
```diff
@@ -101,6 +101,47 @@ func TestKubernetesAndUpgradeOverlay(t *testing.T) {
...
+		return
+	}
+
+	// install
+	if !e.InstallSpinnaker(ns, spinOverlay, t) {
+		return
+	}
+
+	v := RunCommandAndAssert(fmt.Sprintf("%s -n %s get spinsvc %s -o=jsonpath='{.status.status}'", e.KubectlPrefix(), ns, SpinServiceName), t)
+	if t.Failed() || !assert.Equal(t, "OK", strings.TrimSpace(v)) {
```
dogonthehorizon, line 126: Out of curiosity, do we use an enum or similar type for SpinnakerService statuses in Operator?

integration_tests/integration_test.go
```diff
@@ -101,6 +101,47 @@ func TestKubernetesAndUpgradeOverlay(t *testing.T) {
...
+		return
+	}
+
+	if !e.InstallSpinnaker(ns, "testdata/spinnaker/overlay_spinsvc_status", t) {
+		return
+	}
+	time.Sleep(20 * time.Second)
```
dogonthehorizon, line 133: Is it possible for the test to fail if we don't see a status change in 20 seconds? What if we did something like an exponential backoff check to see if the status failed?

integration_tests/integration_test.go
dogonthehorizon: Oh, sorta, looks like it's defined in `spinnakerservice/status.go`, what if we used that here? I think the assertion should still hold

pkg/controller/spinnakerservice/status.go
```diff
@@ -25,81 +27,129 @@ const (
...
-	return s.client.Status().Update(context.Background(), svc)
-}
-
-// getSpinnakerServiceImageFromDeployment returns the name of the image
-func (s *statusChecker) getSpinnakerServiceImageFromDeployment(p v1.PodSpec) string {
-	for _, c := range p.Containers {
-		if strings.HasPrefix(c.Name, "spin-") {
-			return c.Image
+			break
+		case v1.PodFailed, v1.PodUnknown:
```
dogonthehorizon, line 147: This is neat, I didn't know you could collapse case statements like this!

pkg/util/k8s_lookup.go
```diff
@@ -0,0 +1,112 @@
...
+		return []appsv1.Deployment{}, nil
+	} else {
+		return list.Items, nil
+	}
+}
+
+// getSpinnakerServiceImageFromDeployment returns the name of the image
+func (l K8sLookup) GetSpinnakerServiceImageFromDeployment(p v1.PodSpec) string {
+	for _, c := range p.Containers {
+		if strings.HasPrefix(c.Name, "spin-") {
```
dogonthehorizon, line 53: Under what circumstances would the container name start with `spin-`? The few clusters I've been looking at today all have the name of the deployment (e.g. `deploy/spin-front50` has a container named `front50`)

pkg/util/k8s_lookup.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+	// Default to first container if it exists
+	if len(p.Containers) > 0 {
+		return p.Containers[0].Image
+	}
+	return ""
+}
+
+// isContainerInFailureState validate if container is in a failure state
```
dogonthehorizon, line 64: This comment doesn't seem to match the method, is that intentional?

cristhian-castaneda: that was not intentional, I wrote that comment at the first approach, will change it.

pkg/controller/spinnakerservice/status.go
cristhian-castaneda: 😃

integration_tests/integration_test.go
cristhian-castaneda: sure thing, let me do that.

integration_tests/integration_test.go
cristhian-castaneda: that a great Idea will implement an exponential backoff check.

pkg/util/k8s_lookup.go
cristhian-castaneda: To be honest, I don’t know but think we should keep it as is while we get the reason for this check.

dogonthehorizon: np, we'll keep an eye on it!

dogonthehorizon (approved):

PR #179, opened 2020-09-22
fix(test): fix integration tests

cristhian-castaneda:


integration_tests/integration_test.go
```diff
@@ -125,21 +125,24 @@ func TestUpdateSpinsvcStatus(t *testing.T) {
 
 	sc := func() error {
 		v := RunCommandAndAssert(fmt.Sprintf("%s -n %s get spinsvc %s -o=jsonpath='{.status.status}'", e.KubectlPrefix(), ns, SpinServiceName), t)
-		if t.Failed() || !assert.Equal(t, spinnakerservice.Ok, strings.TrimSpace(v)) {
```
cristhian-castaneda, line 128: We need this to not rely on the first status check because when the statement is not fulfilled with the assert function an error is thrown.

integration_tests/integration_test.go
```diff
@@ -125,21 +125,24 @@ func TestUpdateSpinsvcStatus(t *testing.T) {
...
+		if t.Failed() || spinnakerservice.Ok != strings.TrimSpace(v) {
 			return fmt.Errorf("spinnaker is not in %s status yet", spinnakerservice.Ok)
 		}
 
 		return nil
 	}
 	ExponentialBackOff(sc, 3)
 
-	if !e.InstallSpinnaker(ns, "testdata/spinnaker/overlay_spinsvc_status", t) {
+	LogMainStep(t, "Installing spinnaker in namespace %s", ns)
```
cristhian-castaneda, line 136: Same here, we are testing a bad status, SpinnakerService never get OK status

dogonthehorizon (approved):

PR #189, opened 2020-10-19
feat(unmanage): Unmanage spinnaker services

edgarulg:
This change is about provide a way to specify Operator to not manage certain spinnaker services.

it defines a new properties field in the CRD named Operator, it will contain configurations for Operator in this case, a list of services that won't be manage by Operator.

```yml
  operator:
    unmanagedServices:
      - orca
      - echo
```

it skips the update for the `Service` & `Deployment` for the services in the `unmanagedService` list after generating the manifest configs from Halyard.

### DIFF
--- a/pkg/deploy/spindeploy/save.go
+++ b/pkg/deploy/spindeploy/save.go
@@ -18,7 +18,7 @@ import (
 )
 
 // TransformManifests adjusts settings to the configuration
-func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig, logger logr.Logger) error {
+func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig, logger logr.Logger, unmanage map[string]struct{}) error {
 	// Set SpinnakerService instance as the owner and controller
 	count := 0
 	for _, v := range gen.Config {
@@ -36,6 +36,10 @@ func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen
 	// the status. But things happen.
 	d.log.Info(fmt.Sprintf("saving %d manifests across %d services", count, len(gen.Config)))
 	for k := range gen.Config {
+		if _, ok := unmanage[k]; ok {
+			logger.Info(fmt.Sprintf("skip applying updates for %s", k))
+			continue
+		}
 		s := gen.Config[k]
 		if s.Deployment != nil {
 			logger.Info(fmt.Sprintf("saving deployment manifest for %s", k))
--- a/pkg/deploy/spindeploy/save.go
+++ b/pkg/deploy/spindeploy/save.go
@@ -18,7 +18,7 @@ import (
 )
 
 // TransformManifests adjusts settings to the configuration
-func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig, logger logr.Logger, unmanage map[string]struct{}) error {
+func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig, logger logr.Logger, unmanage map[string]string) error {
 	// Set SpinnakerService instance as the owner and controller
 	count := 0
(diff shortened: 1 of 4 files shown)


pkg/deploy/spindeploy/spindeploy.go
```diff
@@ -122,7 +122,14 @@ func (d *Deployer) Deploy(ctx context.Context, svc interfaces.SpinnakerService,
 		}
 	}
 
-	if err = d.deployConfig(ctx, scheme, l, rLogger); err != nil {
+	sUnmanaged := nSvc.GetOperatorConfig().UnmanagedServices
```
edgarulg, line 125: When the reconcile cycle identify a change, it gets the list of services that won't be managed by operator. create a map with the list of name services

pkg/deploy/spindeploy/save.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // TransformManifests adjusts settings to the configuration
-func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig, logger logr.Logger) error {
+func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen *generated.SpinnakerGeneratedConfig, logger logr.Logger, unmanage map[string]string) error {
```
edgarulg, line 21: deployConfig method is responsible of patching the updates to k8s

pkg/deploy/spindeploy/save.go
```diff
@@ -36,6 +36,10 @@ func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen
 	// the status. But things happen.
 	d.log.Info(fmt.Sprintf("saving %d manifests across %d services", count, len(gen.Config)))
 	for k := range gen.Config {
+		if _, ok := unmanage[k]; ok {
```
edgarulg, line 39: when saving manifests for each service, it search the service on the unmanage map, if present it skips save object on k8s (applying)

pkg/deploy/spindeploy/save.go
```diff
@@ -36,6 +36,10 @@ func (d *Deployer) deployConfig(ctx context.Context, scheme *runtime.Scheme, gen
...
+			logger.Info(fmt.Sprintf("skip applying updates for %s", k))
+			continue
+		}
 		s := gen.Config[k]
 		if s.Deployment != nil {
```
edgarulg, line 44: it would skip only the Deployment for the given services adding the condition here

pkg/deploy/spindeploy/save.go
edgarulg: It will skip applying the Deployment & Services for the given services. The resources has a concatenation for the service, so they will be applied.

german-muzquiz (commented):
What is the use case of having services not managed by operator?

edgarulg:
Hi @german-muzquiz, The Spinnaker services are manage by Operator automatically but we would like to have CD for every spin service to test latest changes in a staging env. Spinnakers manage by Operator won't be able to have pipelines to deploy its own service because Operator will reconcile those changes and will try to match the deployment.

Here is a link with a little bit more context: https://armory.atlassian.net/browse/BOB-367

Using this implementation, we can let operators decide if they want Operator manage all the services or just a few of them to deploy the service in a different way.

german-muzquiz:
I think it sounds reasonable to not manage some things about the deployment of spinnaker with the operator. It could be expressed in more general terms.

How about a section that is similar to `deploymentEnvironment` of the hal config, indicating details of how spinnaker will be deployed. Maybe something like this:

```
apiVersion: spinnaker.io/v1alpha2
kind: SpinnakerService
metadata:
  name: spinnaker
spec:
  deployment:
    managementType:
      orca: None
```

managementType options:
- Full (default, current behavior)
- OnlyConfigs (only generates and deploys secrets and configmaps for that service, but excludes service and deployment objects)
- None (doesn't generate or deploy anything for that service)

The reason is that since we're moving away of Halyard we need a way to generate the base manifest files of the services. `spec.deployment` could include any additional field that we need for that. The option `OnlyConfigs` is compatible with what kleat does.

I'd like @ncknt input as well for this PR.

edgarulg:
I am going to close this PR because I will move this feature to different project. This will help to keep operator consisten on its configuration.

Issue #204, opened 2021-01-11
kustomise: doesn't work with service port definition


georgestack:
Hello,

We are using version:
```  
spinnakerVersion: 1.19.4
  image:
    repository: gcr.io/spinnaker-marketplace/halyard
    tag: 1.32.0
```

We have noticed that when using `kustomise:` for spin-gate service, changes made on the ports: section of the service are not effective.

With below configuration it is expected that another port named test would appear in the service. 
```
      spec:
        kustomize:
          gate:
            service:
              patches:
                - |
                  metadata:
                    labels:
                      test: test
                  spec:
                    ports:
                    - name: test
                      port: 8085
                      protocol: TCP
                      targetPort: 8084
                    selector:
                      test: spinnaker
```

As a result we get:
```
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2021-01-11T11:08:46Z"
  labels:
    app: spin
    cluster: spin-gate
    prometheus: spin-gate
  name: spin-gate
  namespace: spinnaker
  ownerReferences:
  - apiVersion: spinnaker.io/v1alpha2
    blockOwnerDeletion: true
    controller: true
    kind: SpinnakerService
    name: spinnaker
  resourceVersion: "1302293"
  selfLink: /api/v1/namespaces/spinnaker/services/spin-gate
spec:
  clusterIP: 172.20.52.42
  ports:
  - name: gate-tcp
    port: 443
    protocol: TCP
    targetPort: 8084
  selector:
    app: spin
    cluster: spin-gate
    test: spinnaker
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```

even though changes using kustomize: are effective in other sections of the service like selector or labels ports are not updated. Noticed in the source code and from tested behavior that no matter what we add as port in kustomize: the ports in the service are taken by overrideBaseUrl parameter, if left empty they are chosen based on protocol (443/80):
```
apiSecurity:
    overrideBaseUrl: 
```

I believe this is a conflict of kustomize support and code, can you confirm if this is intended behavior? Why it should not be possible to add secondary service port through kustomize?

man0s:
I was having a similar problem, I have opened the following PR: https://github.com/armory/spinnaker-operator/pull/208

cristhian-castaneda:
@georgestack thanks for the report, and apologies for the slow response.

[...]

PR #208, opened 2021-01-26
fix(interface): add PublicPort check

man0s:
**Why:**
If `aUrl` is defined without port `u.Scheme` will override the `defaultPort` even if it is defined in `expose.service.overrides.{svc}.publicPort`.

**Fixes:**
- https://github.com/armory/spinnaker-operator/issues/204


cristhian-castaneda:
@man0s thanks for your contribution, there is an open PR that tackle this issue https://github.com/armory/spinnaker-operator/pull/210.

PR #227, opened 2021-04-26
feat(lambda/validation): Validatation regarding the AWS Lambda using the GO SDK to get the Lambda Functions using the AWS provider credentials

jorgebee65:
Validation for AWS Lambda.
Using the GO SDK to get the Lambda Functions using the AWS provider credentials. (aws-sdk-go)

In order to pass the validation is necessary to have an AWS provider enabled and at least with the SecretKeyI, AccesKeyId, defaultRegions configured, and enable Spinnaker to deploy apps to AWS Lambda:

```
apiVersion: spinnaker.io/v1alpha2
kind: SpinnakerService
metadata:
  name: test
spec:
  spinnakerConfig:
    config:
      providers:
        aws:
          accessKeyId: IHD9A8SA9D8HAS9
          secretAccessKey: secret12345678
          enabled: true
          defaultRegions:
          - name: us-west-2
    profiles:
      clouddriver:
        aws:
          accounts:
            - accountId: "569630529054"
              assumeRole: role/spinManaged-test
              lambdaEnabled: true
              name: aws-ps-dev
          features:
            lambda:
              enabled: true
      deck:
        settings-local.js: |
          window.spinnakerSettings.feature.functions = true
  validation:
    providers:
      lambda:
        enabled: true
```


pkg/validate/lambda_validator.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/armory/spinnaker-operator/pkg/apis/spinnaker/interfaces"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/lambda"
+	"os"
+	"strconv"
+)
+
+const (
+	lambdaAccountType        		= "lambda"
```
rvazquezglez, line 14: Nitpicking, here is a mix of tabs and spaces, I think it should be either spaces or tab.

pkg/validate/lambda_validator.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return ValidationResult{Errors: []error{fmt.Errorf("SecretAccessKey is missing")}}
+	}
+
+	//Lambda
+	isEnabled, err := spinSvc.GetSpinnakerConfig().GetRawServiceConfigPropString("clouddriver",lambdaClouddriverEnabledKey)
+	if err != nil {
+		return ValidationResult{}
+	}
+	accountEnabled, err := strconv.ParseBool(isEnabled)
```
rvazquezglez, line 52: Looks like this `err` is ignored, is that on purpose?

pkg/validate/lambda_validator.go
```diff
@@ -0,0 +1,91 @@
...
+	os.Setenv("AWS_SECRET_ACCESS_KEY", secretKey)
+
+	conf := aws.Config{Region: aws.String(region)}
+	sess := session.New(&conf)
+	svc := lambda.New(sess)
+	input := &lambda.ListFunctionsInput{}
+
+	_, err := svc.ListFunctions(input)
+	if err != nil {
+    	return false, []error{fmt.Errorf(err.Error())}
```
rvazquezglez, line 87: There might be a missing indentation here.

zachsmith1:
I think we also need to add the LambdaEnabled property in Halyard because otherwise halyard will still fail this while trying to deploy it ( I think )

pkg/validate/lambda_validator.go
jorgebee65: fixed

pkg/validate/lambda_validator.go
jorgebee65: The purpose of the err only happen if the property **aws.features.lambda.enabled** is missing: that means it is not enabled and then return no errors

pkg/validate/lambda_validator.go
jorgebee65: Fixed

jorgebee65:
> I think we also need to add the LambdaEnabled property in Halyard because otherwise halyard will still fail this while trying to deploy it ( I think )

The **lambdaEnabled** attribute should be configured under:     
```
    profiles:  
      clouddriver:                         
        aws:
          accounts:
          - name: aws-ps-dev
            lambdaEnabled: true
```
This way halyard will not complain about the attribute, if is colocated in other place then halyard will fail.

zachsmith1 (approved):
LGTM

PR #236, opened 2021-06-02
fix(it): fix integration tests

cristhian-castaneda:
This PR enables us to run the integration tests successfully.
We are not longer using EKS infrastructure to perform these tests.

We are now using kind (https://kind.sigs.k8s.io/), as this helps us to have an ephemeral k8s cluster for these tests.


integration_tests/env.go
```diff
@@ -15,22 +15,25 @@ import (
...
-	OperatorImageVar = "OPERATOR_IMAGE"
-	HalyardImageVar  = "HALYARD_IMAGE"
-	BucketVar        = "S3_BUCKET"
-	BucketRegionVar  = "S3_BUCKET_REGION"
-
+	KubeconfigVar                        = "KUBECONFIG"
+	OperatorImageVar                     = "OPERATOR_IMAGE"
+	HalyardImageVar                      = "HALYARD_IMAGE"
+	BucketVar                            = "S3_BUCKET"
+	BucketRegionVar                      = "AWS_DEFAULT_REGION"
```
cristhian-castaneda, line 22: We'll rely on an IAM user to use the bucket we provide for tests

integration_tests/integration_test.go
```diff
@@ -24,7 +24,16 @@ func init() {
 	}
 }
 
-func TestSpinnakerBase(t *testing.T) {
+func TestIntegrationTests(t *testing.T) {
+	t.Run("Operator=1", SpinnakerBase)
```
cristhian-castaneda, line 28: We are now executing the tests sequentially, since we are using a local cluster it is good to not overload it.

integration_tests/testdata/operator/base/deployment-template.yaml
```diff
@@ -18,17 +18,30 @@ spec:
           image: {{.OperatorImage}}
           command:
             - spinnaker-operator
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
cristhian-castaneda, line 21: Since we are loading the operator image within kind we are not needing to reload it.

integration_tests/utils.go
```diff
@@ -124,7 +123,7 @@ func WaitForLBReady(ns, statusPath string, e *TestEnv, t *testing.T) string {
 			lbUrl, _ = RunCommandSilent(fmt.Sprintf("%s -n %s get spinsvc %s -o=jsonpath='%s'", e.KubectlPrefix(), ns, SpinServiceName, statusPath), t)
 		}
 		if lbUrl != "" {
-			_, err := RunCommandSilent(fmt.Sprintf("curl %s", lbUrl), t)
+			_, err := RunCommandSilent(fmt.Sprintf("docker run --rm --network host curlimages/curl %s", lbUrl), t)
```
cristhian-castaneda, line 126: Since the integration tests are interacting with the spinnaker URL and the URL is generated within the docker network.
We need to emulate a curl call inside of the docker network, so we are using a docker image instead of the CURL command itself for it.

integration_tests/utils.go
```diff
@@ -290,3 +284,16 @@ func ExponentialBackOff(operation backoff.Operation, minutes time.Duration) erro
 
 	return backoff.Retry(operation, b)
 }
+
+func GetLocalHost(t *testing.T) string {
```
cristhian-castaneda, line 288: The kubeconfig that is used to spin up the operator is the same that is used to create a Kubernetes account on spinnaker, so, we need to change the cluster URL.

integration_tests/env.go
dogonthehorizon: What permissions does the user have, and how is it created?

integration_tests/env.go
```diff
@@ -122,20 +127,33 @@ func resolveEnvVars(d Defaults, t *testing.T) Vars {
...
 	r := os.Getenv(BucketRegionVar)
 	if r == "" {
 		t.Logf("%s env var not set, using default", d.BucketRegionDefault)
 		r = d.BucketRegionDefault
 	}
 	t.Logf("Using bucekt region %s", r)
+
+	a := os.Getenv(BucketAccessKeyIdVar)
+	if r == "" {
+		t.Logf("%s env var not set, using default", d.BucketRegionDefault)
```
dogonthehorizon, line 141: I believe this log message needs to be updated with the access key id var?

integration_tests/env.go
```diff
@@ -122,20 +127,33 @@ func resolveEnvVars(d Defaults, t *testing.T) Vars {
...
+		r = d.BucketRegionDefault
+	}
+	s := os.Getenv(BucketSecretAccessKeyVar)
+	if r == "" {
+		t.Logf("%s env var not set, using default", d.BucketRegionDefault)
```
dogonthehorizon, line 146: I believe this log message needs to be updated with the secret access key var?

I think both of these checks need to fail as well. Presumably if we don't have permissions into AWS then we can't run our integration tests?

integration_tests/env.go
```diff
@@ -170,6 +188,7 @@ func InstallCrdsAndOperator(spinNs string, isClusterMode bool, d Defaults, t *te
 
 func (e *TestEnv) InstallCrds(d Defaults, t *testing.T) bool {
 	ApplyManifest("default", d.CRDManifests, e, t)
+	_, _ = RunCommand("rm -rf ~/.kube/http-cache/ && rm -rf ~/.kube/cache/", t)
```
dogonthehorizon, line 191: Are there any consequences to this command failing?

integration_tests/integration_test.go
dogonthehorizon: How much does that impact overall integration test times?

integration_tests/kind/metallb-configmap.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  namespace: metallb-system
+  name: config
+data:
+  config: |
+    address-pools:
+    - name: default
+      protocol: layer2
+      addresses:
+      - 172.17.255.1-172.17.255.250
```
dogonthehorizon, line 12: Is this the IP range that `kind` uses? If so can we state where that's documented in a comment?

integration_tests/utils.go
dogonthehorizon: This would be great as a comment in the codebase! Very useful context.

integration_tests/env.go
cristhian-castaneda: The account that we are using here was the one that was created when we spin-up the integration-test cluster on the dev account, it was created with terraform. should we create a new one  🤷🏼‍♂️?
if we merge this change the current infrastructure for it will not be needed.

integration_tests/env.go
cristhian-castaneda: Yeah, you are right we need to have the AWS credentials in order to make tests work, will mark those as a `Faltaf`

integration_tests/env.go
cristhian-castaneda: Not necessarily, I'm mainly using it to let `kubectl`  go and check the CRD on the cluster itself instead of the local cache.
I realize that if I don’t do that, kubectl doesn’t recognize the `spinsvc`, `spinnakeraccounts` because it gets just information from the cache.
So, maybe a warning message?

integration_tests/integration_test.go
cristhian-castaneda: It does impact substantially, the normal execution timeline on the current execution on the spinnaker pipeline is roughly 15 - 20 minutes.
Not it could take up to 40 minutes.

integration_tests/kind/metallb-configmap.yaml
cristhian-castaneda: yup, it is the IP range that `kind` will use to expose load balancer services.
I just added a comment for it.

integration_tests/utils.go
cristhian-castaneda: I just added

integration_tests/integration_test.go
dogonthehorizon: Okay, that does sound like quite a bit. Let's keep an eye on PR times and see if we need to reconsider this change.

dogonthehorizon (approved):

Issue #245, opened 2021-09-01
Operator projects using the removed APIs in k8s 1.22 requires changes


camilamacedo86:
Based on the comment. It should be here: https://github.com/spinnaker/spinnaker/issues/6520#issuecomment-910729798
### Problem Description

Kubernetes has been deprecating API(s), which will be removed and are no longer available in 1.22. Operators projects using [these APIs](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22) versions will **not** work on Kubernetes 1.22 or any cluster vendor using this Kubernetes version(1.22), such as OpenShift 4.9+. Following the APIs that are most likely your projects to be affected by:
- apiextensions.k8s.io/v1beta1: (Used for CRDs and available since v1.16)
- rbac.authorization.k8s.io/v1beta1: (Used for RBAC/rules and available since v1.8)
- admissionregistration.k8s.io/v1beta1 (Used for Webhooks and available since v1.16)

Therefore, looks like this project distributes solutions in the [repository](https://github.com/redhat-openshift-ecosystem/community-operators-prod/tree/main/operators/spinnaker-operator) and does not contain any version compatible with k8s 1.22/OCP 4.9. ([More info](https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/138)). Following some findings by checking the distributions published:

- **spinnaker-operator.v1.21.2**: this distribution is using APIs which were deprecated and removed in v1.22. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. Migrate the API(s) for CRD: (["spinnakeroperators.charts.helm.k8s.io"]) 
- **spinnaker-operator.v1.13.1**: this distribution is using APIs which were deprecated and removed in v1.22. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. Migrate the API(s) for CRD: (["spinnakeroperators.charts.helm.k8s.io"]) 
- **spinnaker-operator.v1.17.2**: this distribution is using APIs which were deprecated and removed in v1.22. More info: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. Migrate the API(s) for CRD: (["spinnakeroperators.charts.helm.k8s.io"]) 
[...]

PR #250, opened 2022-02-11
Changes the yaml for crd creation to remove deprecation warning on v1.22

seanrclayton:
I ran into the same problem described in [this issue](https://github.com/armory/spinnaker-operator/issues/245) I updated the yaml to comply with the changes described in the migration document [here](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22) And Deploying the service seems to work just fine.

### DIFF
--- a/deploy/crds/spinnaker.io_spinnakeraccounts_crd.yaml
+++ b/deploy/crds/spinnaker.io_spinnakeraccounts_crd.yaml
@@ -1,21 +1,454 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: spinnakeraccounts.spinnaker.io
 spec:
-  additionalPrinterColumns:
-  - JSONPath: .spec.type
-    description: Type
-    name: type
-    type: string
-  - JSONPath: .status.LastValidatedAt
-    description: Last Validated
-    name: lastValidated
-    type: date
-  - JSONPath: .status.InvalidReason
-    description: Invalid Reason
-    name: reason
-    type: string
+
+  versions:
+  - name: v1alpha2
+    additionalPrinterColumns:
+    - jsonPath: .spec.type
+      description: Type
+      name: type
+      type: string
+    - jsonPath: .status.LastValidatedAt
+      description: Last Validated
+      name: lastValidated
+      type: date
+    - jsonPath: .status.InvalidReason
+      description: Invalid Reason
+      name: reason
+      type: string
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema: 
+      openAPIV3Schema:
+        description: SpinnakerAccount is the Schema for the spinnakeraccounts API
+        properties:
+          apiVersion:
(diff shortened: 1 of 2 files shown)


seanrclayton:
Apologies - I didn't realize i hadn't merged the feature branch with master in my repo. I actually changed my original way of fixing and issue with the depreciation of  `anyOf` to use [x-kubernetes-int-or-string](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#intorstring) The CRD applies now. I tried testing it on my local machine rather than a cluster - but got an error, so further testing on spinnaker's end would be needed.

villasenor:
@seanrclayton Thanks for jumping in to help solve this issue! Not sure if this is due to an error on my end or the changes in this branch, but while I was able to get the operator to install with this branch, I can't actually get the operator to deploy anything. See the issue I opened here: https://github.com/armory/spinnaker-operator/issues/251

deploy/crds/spinnaker.io_spinnakeraccounts_crd.yaml
```diff
@@ -1,460 +1,456 @@
...
+                  enabled:
+                    description: Enable or disable validation, defaults to false
+                    type: boolean
+                  failOnError:
+                    description: Report errors but do not fail validation, defaults
+                      to true
+                    type: boolean
+                  frequencySeconds:
+                    x-kubernetes-int-or-string: true
+                    description: Number of seconds between each validation
```
abannachZen, line 416: I found that you can still use the `anyOf` syntax, but the order of the types matters. 
In the old version it was 
```
anyOf:
  - type: string
  - type: integer
```  
but 
```
anyOf:
  - type: integer
  - type: string
```  
avoids the need to use `x-kubernetes-int-or-string`. My interpretation of `x-kubernetes-int-or-string` was a workaround, but that may be misguided.

```suggestion
                  frequencySeconds:
                    type: object
                    anyOf:
                      - type: integer
                      - type: string
                    description: Number of seconds between each validation
```

abannachZen: The above suggestion can be applied to all `int-or-string` lines.

abannachZen: Reading more of the documentation looks like `x-kubernetes-int-or-string` is just shorthand, not necessarily a workaround. Carry on!

ArtemChekunov:
Hi guys,

Many thanks for all your work.
Any chance to release it before GKE will upgrade to kube_ver: 1.22?

ArtemChekunov:
hmm
> extra keys not allowed @ pull_request_rules → item 0 → actions → merge → strict

looks like the field is not appropriate https://github.com/armory/spinnaker-operator/blob/master/.mergify.yml#L10

Eugst:
the PR need to be approved, that's why mergify is erroring. 

@seanrclayton could you, please, check the comment from @abannachZen

Eugst:
@abannachZen How I can speed up the delivery of this PR?

evin-bz:
I can confirm this fixes the issue I was having as well (with k3d), would be a +1 to merging

Issue #253, opened 2022-04-19
Mergify: strict mode deprecated


Eugst:
https://blog.mergify.com/strict-mode-deprecation/ strict mode was deprecated.
mergify config need to be refactored.

Eugst:
#254

dogonthehorizon:
@Eugst it looks like we have resolved this in another commit that exists on master. Thank you for the contribution, but I'll go ahead and close this (and the related PR) for now unless we need further changes made to the mergify config!

PR #254, opened 2022-04-19
fixed mergify deprecated config.

Eugst:


Eugst:
https://github.com/armory/spinnaker-operator/issues/253

Eugst:
@ncknt @andrewbackes @DanJohnstonArmory could you help me with this?
I would like to unlock and speedup #250

dogonthehorizon:
@Eugst it looks like we have resolve this in another commit that exists on master. Thank you for the contribution, but I'll go ahead and close this (and the related issue) for now unless we need further changes made to the mergify config!

PR #266, opened 2022-08-16
Support building both amd64 and arm64 docker images

mflis:
- docker buildx doesn't allow to easily separate build and push step
so `make docker-package` and `make docker-push` needed to be merged
into one command

- Using locally built image in FROM directive seems impossible
with `docker buildx` - remote docker registry is required.
So all dockerfiles were merged into one.
See: https://github.com/docker/buildx/issues/301#issuecomment-755164475

- AWS_AIM_AUTHENTICATOR_VERSION and GOOGLE_CLOUD_SDK_VERSION were bumped
  to versions where arm64 binaries are available


mflis:
Build failed on "Login to docker registry", because it's forked repository. @ichi0915  could you help here? Also I see that integration tests were failing for last few merges to master, so it will probably fail on my MR too.

ichi0915:
Yeah it is failing because it is a fork we are going to work on that after we fix the integration tests.

The other day I tested this PR and it didn't do the push to docker I just saw that you updated the makefile and it looks good, I'm going to do some more test with the newer changes.

FYI I cannot merge these changes because I don't have permissions to merge until the actions have passed, I'm going to tell someone in armory to merge this but first we are merging this other [PR](https://github.com/armory/spinnaker-operator/pull/267)

PR #282, opened 2023-02-13
chore(release): update deployment manifest with specific release tag

ciurescuraul:
For the OSS operator release bundle, the deployment manifest always points to the latest image instead of pointing to specific tags for the respective Docker images.

Relates to [BOB-30954](https://armory.atlassian.net/browse/BOB-30954)


ovidiupopa07 (approved):
LGTM

PR #287, opened 2023-03-07
feat(k8s): add support for GA ingress apiVersion in helm chart and operator

nasonfish:
Hi there,

We're currently noticing that this spinnaker operator creates an `Ingress` object of version `extensions/v1beta1`. This version is no longer supported by versions of Kubernetes v1.22+, as per [this documentation](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122).

This change implements what `helm create` templates out by default for Ingresses, allowing backwards-compatibility for folks using the old ingress version while allowing installation of the GA Ingress `apiVersion` (conditional on `.Capabilities.KubeVersion.GitVersion`).

In short: this should be a backwards-compatible change which allows installation of this operator on clusters running Kubernetes v1.22+.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,5 @@ integration_tests/testdata/spinnaker/overlay_secrets/kubecfg
 integration_tests/testdata/**/*-generated.yml
 integration_tests/testdata/**/*-generated.yaml
 hack/tools/bin/controller-gen
+
+vendor/
--- a/.gitignore
+++ b/.gitignore
@@ -86,5 +86,3 @@ integration_tests/testdata/spinnaker/overlay_secrets/kubecfg
 integration_tests/testdata/**/*-generated.yml
 integration_tests/testdata/**/*-generated.yaml
 hack/tools/bin/controller-gen
-
-vendor/
--- a/deploy/operator/helm/templates/ingress.yaml
+++ b/deploy/operator/helm/templates/ingress.yaml
@@ -1,18 +1,7 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "spinnaker-operator.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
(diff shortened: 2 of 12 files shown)


deploy/operator/helm/templates/ingress.yaml
```diff
@@ -1,17 +1,32 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "spinnaker-operator.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
+{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
```
jasonmcintosh, line 4: ONE thing I've not tested - i thought capabilities was a hook and... if you're using spinnaker to try to do this it MAY not work...

nasonfish: Hmm, good point. I think we've seen this before as well.

How would you feel about me just solidly switching over to the GA Ingress version, and then updating the compatibility matrix to say that whatever version this ends up in does not support kubernetes <`1.19`?

jasonmcintosh: Note, I'd just replace this entirely and remove the older non supported ingress versions ;) AND state that we only support k8s greater than 1.18.

jasonmcintosh: 100% on "don't support k8s older than 1.19 ;) If that level of support is needed, the older releases work well.

nasonfish:
hey there @jasonmcintosh -- you are absolutely right, this was missing some essential upgrades to the actual operator. It has now been updated to use the GA ingress version. it appears to be running successfully -- let me know if you have a chance to take a look at this again.

mrampton:
Thanks for the work on this @nasonfish!

.gitignore
```diff
@@ -86,3 +86,5 @@ integration_tests/testdata/spinnaker/overlay_secrets/kubecfg
 integration_tests/testdata/**/*-generated.yml
 integration_tests/testdata/**/*-generated.yaml
 hack/tools/bin/controller-gen
+
+vendor/
```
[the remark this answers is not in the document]
nasonfish: done, sorry about that!

PR #304, opened 2025-09-19
fix(AWSCreds): Caused by race condition while processing s3 artifact

tonyg4864:
* Fixed issue where AWS secret env variable was missing from clouddriver components when running in HA mode. Specifically when enabling s3 artifact credentials.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,118 @@
+# v1.4.0
+
+## Features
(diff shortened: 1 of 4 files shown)


jasonmcintosh (approved):
I'll note:  This is pretty much "deprecated" at this point in favor of kustomize, but this still looks like a nice fix!

